PR #7, opened 2014-07-29
Make ParserUtil Timeouts configurable

swoker:


schildbach:
The static ParserUtils class is now migrated into HttpClient instances. So we could configure timeouts as proper instance variables, if this is still wanted/needed. If so, feel free to open a new PR, rebased to current master.

PR #24, opened 2015-02-01
Navitia: Fix queryTrips and query context builder.

### LINKED ISSUE #18: ParisProvider: queryTrips returns only 1 trip
I queried a couple of trips and there's usually only one trip returned while all the other providers return at least 3 trips.

Also, querying for earlier or later trips does not work for me, at least not the same way it does for other providers.

This is probably for @aelkhour.

aelkhour:
This should fix #18


schildbach:
Merged.

PR #30, opened 2015-02-25
Navitia: Fix Location type handling in queryNearbyLocations.

### LINKED ISSUE #26: Navitia: queryNearbyLocations requires location.isIdentified()
@aelkhour In [AbstractNavitiaProvider.java:763](https://github.com/schildbach/public-transport-enabler/blob/master/enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java#L763) you are using location.isIdentified() instead of location.hasLocation() which breaks querying for locations with LocationType.ANY that only have GPS coordinates.

Other AbstractProviders use location.hasLocation() as well, so I suggest using this here, too. But @schildbach would know that better than me.

aelkhour:
This should fix #26


schildbach:
Merged, thanks.

PR #32, opened 2015-03-13
More bugfixes for VRS Provider

mdyrna:
I found and fixed another two bugs:
* Product filter for CABLECAR generated syntactically incorrect web API call, which lead to an unhandled JSON error code and Öffi crash.
* Query for earlier trips returned trips much too early because arrival limit was not set to earliest previous trip arrival but earliest previous segment arrival.

Please pull before merging to your master branch.


schildbach:
Thanks! I merged this to my provider-vrs branch.

PR #34, opened 2015-03-18
Bugfixes and Improvements

mdyrna:
* empty polygon in queryTrips API result caused NumberFormatException
* queryNearbyLocations passed swapped latitude and longitude to web API
* added styles for several Cologne bus lines (by far not all)


schildbach:
Thanks! I merged the queryNearbyLocations fix and bus line styles.

As for the rest of the PR, there are two @Nullable errors. You'll probably need to configure your JSR305 annotations so you can see the errors.

In future, can you separate different bugfixes or features at least by using separate commits? (You could even use separate PRs but for now that's up to you.)

mdyrna:
The two @Nullable errors are fixed now although they were "only" in the test class.

schildbach:
I merged the rest of the PR, after changing usage of the deprecated Date constructor to using Iso8601Format.

PR #41, opened 2015-07-16
Realtime Departures for MVV (Muenchen)

mdyrna:
In addition to departures from the EFA interface it queries the Google Web Toolkit Remote Procedure Call (GWT-RPC) interface used by the app "MVG Fahrinfo" and merges the predicted times into the EFA departures.

Reasons for merging instead of separate departures-only provider:
- The GWT-RPC interface know only the predicted times, not the planned times. So the Öffi UI wouldn't be able to display the delta ("+2", etc.)
- MVG considers the support for suburban trains (S-Bahn) experimental. (The MVG Fahrinfo App doesn't even show S-Bahn departures.)
- The interface doesn't know the regional trains of Bayerische Oberlandbahn (BOB) and Meridian, which are tightly integrated into the Munich S-Bahn network.
- The number of departures is limited and cannot be set in the query.

### DIFF
--- a/enabler/src/de/schildbach/pte/MvvRtProvider.java
+++ b/enabler/src/de/schildbach/pte/MvvRtProvider.java
@@ -1,0 +1,238 @@
+package de.schildbach.pte;
+
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import javax.annotation.Nullable;
+
+import com.google.common.base.Charsets;
+import com.google.common.base.Strings;
+
+import de.schildbach.pte.dto.Departure;
+import de.schildbach.pte.dto.Line;
+import de.schildbach.pte.dto.Location;
+import de.schildbach.pte.dto.LocationType;
+import de.schildbach.pte.dto.Product;
+import de.schildbach.pte.dto.QueryDeparturesResult;
+import de.schildbach.pte.dto.ResultHeader;
+import de.schildbach.pte.dto.StationDepartures;
+import de.schildbach.pte.MvvProvider;
+import de.schildbach.pte.util.ParserUtils;
+
+public class MvvRtProvider extends MvvProvider {
+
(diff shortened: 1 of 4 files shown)


schildbach:
I tried building an Öffi version, but it crashed pretty quick:

```
E/AndroidRuntime(15906): java.lang.RuntimeException: uncategorized problem while querying departures on 80012137 de.schildbach.pte.MvvRtProvider@252af2ff
E/AndroidRuntime(15906): 	at de.schildbach.oeffi.stations.QueryDeparturesRunnable.doRequest(QueryDeparturesRunnable.java:110)
E/AndroidRuntime(15906): 	at de.schildbach.oeffi.stations.QueryDeparturesRunnable.run(QueryDeparturesRunnable.java:50)
E/AndroidRuntime(15906): 	at android.os.Handler.handleCallback(Handler.java:739)
E/AndroidRuntime(15906): 	at android.os.Handler.dispatchMessage(Handler.java:95)
E/AndroidRuntime(15906): 	at android.os.Looper.loop(Looper.java:135)
E/AndroidRuntime(15906): 	at android.os.HandlerThread.run(HandlerThread.java:61)
E/AndroidRuntime(15906): Caused by: java.lang.IndexOutOfBoundsException
E/AndroidRuntime(15906): 	at java.util.LinkedList.get(LinkedList.java:519)
E/AndroidRuntime(15906): 	at de.schildbach.pte.MvvRtProvider.queryDepartures(MvvRtProvider.java:187)
E/AndroidRuntime(15906): 	at de.schildbach.oeffi.stations.QueryDeparturesRunnable.doRequest(QueryDeparturesRunnable.java:69)
E/AndroidRuntime(15906): 	... 5 more
```

Can you rebase your branch on current master? You were missing the Maven -> Gradle commit.

Also just as a note, I'm currently refactoring ParserUtils – pulling out HTTP access into a HttpClient class which you can keep for state (cookies etc). It's not a complicated refactoring but it will affect your changes to ParserUtils.

mdyrna:
I pushed a fix. The case that the MvvProvider doesn't return any departures was not handled, thus the IndexOutOfBoundsException. (Which station did you test btw.?)

I branched from the commit right before the Gradle change intentionally because I haven't worked with Gradle before and don't have Eclipse support for it installed yet and wanted to concentrate on the real development work.

schildbach:
I just pushed the HttpClient refactoring and pushed a commit on top that adds support for custom request headers. Do you think you can rebase on that?

Regarding Maven: I think the last build.xml should still work. If you just do "git revert de5db3cb989de2f729d5115dcf197e76888b8fe5" on your branch you should have Maven back. This workaround will not work forever, but it will give you time to migrate.

PR #44, opened 2015-07-25
Refactored chained if statements to switch

andreassimon:


schildbach:
I generally don't like switch blocks and prefer if or if/else blocks instead. What's the reason for this refactoring?

btw. Your nice formatting will not stay like this as soon as we run the formatter.

andreassimon:
I just think in this specific case, switch is more succinct and readable. But that advantage will decay when the formatter breaks the one-liners.
Besides, the switch is probably a little faster: http://stackoverflow.com/questions/6705955/why-switch-is-faster-than-if But this is the least of my arguments.

schildbach:
I just noticed that the if constructs you're refactoring to a switch statement do not exist any more on master. So I'm closing this.

PR #55, opened 2015-12-09
Add support for Navitia fr-sw region (which covers South-West of France)

kalon33:
This commit should add proper initial support for Navitia fr-sw region. Further testing may be needed to check proper tram/bus... logotypes rendering.

### DIFF
--- a/enabler/src/de/schildbach/pte/FrenchSouthWestProvider.java
+++ b/enabler/src/de/schildbach/pte/FrenchSouthWestProvider.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2014-2015 the original author or authors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package de.schildbach.pte;
+
+import de.schildbach.pte.dto.Product;
+import de.schildbach.pte.dto.Style;
+import de.schildbach.pte.dto.Style.Shape;
+
+/**
+ * @author Nicolas Derive
+ */
+public class FrenchSouthWestProvider extends AbstractNavitiaProvider
+{
+	private static String API_REGION = "fr-sw";
+
+	public FrenchSouthWestProvider(final String authorization)
+	{
+		super(NetworkId.ALPC, authorization);
+
+		setTimeZone("Europe/Paris");
+	}
+
+	@Override
+	public String region()
+	{
+		return API_REGION;
+	}
+
+	@Override
+	protected Style getLineStyle(final Product product, final String code, final String color)
+	{
+		switch (product)
+		{
+			case TRAM:
+			{
+				// Tram
+				return new Style(Shape.CIRCLE, Style.TRANSPARENT, Style.parseColor(color), Style.parseColor(color));
+			}
+			case BUS:
+			{
+				// Bus + Transgironde
+				return new Style(Shape.ROUNDED, Style.parseColor(color), computeForegroundColor(color));
+			}
+			case FERRY:
+			{
+				// Batcub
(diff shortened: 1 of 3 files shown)


enabler/src/de/schildbach/pte/FrenchSouthWestProvider.java
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * @author Nicolas Derive
+ */
+public class FrenchSouthWestProvider extends AbstractNavitiaProvider
+{
+	private static String API_REGION = "fr-sw";
+
+	public FrenchSouthWestProvider(final String authorization)
+	{
+		super(NetworkId.ALPC, authorization);
```
schildbach: ALPC? That doesn't match with the new id in NetworkId.java.

schildbach:
Thanks for your contribution! Can you add a FrenchSouthWestProviderLiveTest.java with at least a basic test per query? Have a look at ParisProviderLiveTest.java for an idea.

enabler/src/de/schildbach/pte/FrenchSouthWestProvider.java
kalon33: Sorry, I will fix that, I just renamed the class and Network as it doesn't only covers a french administrative region (Aquitaine-Limousin-Poitou-Charentes).

kalon33:
@schildbach  Is that OK now? I wasn't sure of what certain tests were about.

Thanks for your help :)

kalon33:
Do I need to do more editing or is that good?

enabler/test/de/schildbach/pte/live/FrenchSouthWestProviderLiveTest.java
```diff
@@ -0,0 +1,193 @@
...
+{
+	public FrenchSouthWestProviderLiveTest()
+	{
+		super(new FrenchSouthWestProvider(Secrets.NAVITIA_AUTHORIZATION));
+	}
+
+	@Test
+	public void nearbyStationsAddress() throws Exception
+	{
+		nearbyStationsAddress(44826434, -0557312);
```
schildbach: Literals starting with 0 are interpreted as octal. Are you sure this is what you want to do?

schildbach:
Generally looks good. A couple of unit tests fail, but I realize we have this problem with ParisProvider too currently. Maybe these issues should be fixed separately.

enabler/test/de/schildbach/pte/live/FrenchSouthWestProviderLiveTest.java
kalon33: The relevant place coordinates are 44.826434, -0.557312 respectively. Is this malformed?

kalon33:
@schildbach which ones are failing? Why? I might have misunderstood the roles of some tests. I took ParisProvider as a basis to understand them but if ParisProvider is malformed, I may have reproduced the mistakes inside. 

Could you pull it "as is" and I would fix them after? I would need these changes to be integrated to an app I use, Transportr, that use your project and I will travel to the region covered by my PR next week.

Thanks for your help.

enabler/test/de/schildbach/pte/live/FrenchSouthWestProviderLiveTest.java
schildbach: -0557312 is interpreted as octal so it's -188106 decimal. So your coordinate is -0.188106.

And -0580036 is actually an invalid number (overflows) and fails to compile.

schildbach:
Don't worry, I intend to merge this even with failing tests. But it needs to compile…

kalon33:
@schildbach what's needed to properly compile?

grote:
@kalon33 Look at what Andreas was writing here:
> -0557312 is interpreted as octal so it's -188106 decimal. So your coordinate is -0.188106.
> And -0580036 is actually an invalid number (overflows) and fails to compile.

There he explains, why it does not compile.

kalon33:
@grote so what should I write there to correct this?

grote:
@kalon33 I don't have time at the moment to look into this in detail. Just make sure you are passing proper coordinates to the functions. This might also be helpful to understand what Andreas is referring to: https://stackoverflow.com/questions/7218760/why-is-08-not-a-valid-integer-literal-in-java

kalon33:
So -0557312 should be -.557312 and -0580036 should be -.580036?

schildbach:
Just remove the leading zeroes from integers if you intend them to be interpreted as decimal. I can do it for you if you want.

kalon33:
@schildbach I just did it (this was my fault, so I corrected it). So now, -557312 will be considered as -0.557312 longitude coordinate?

schildbach:
Thanks, I squashed and merged your PR.

PR #58, opened 2015-12-19
Configure continuous integration using Travis CI.

schildbach:
Builds will be visible at https://travis-ci.org/schildbach/public-transport-enabler

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,3 @@
+# configuration for https://travis-ci.org/schildbach/public-transport-enabler
+language: java
+jdk: openjdk6


grote:
Awesome! Any plans to let Travis also run the live tests as well? Maybe they have a nice dashboard to always show the test status of the various providers. Everybody could easily see where fixes might be necessary and subscribe to certain providers.

schildbach:
Hmm, I didn't plan it. Obviously the live tests can't fail the build, otherwise it would be red all the time (-:

A dashboard would be nice, but I have no idea how to do this. Travis itself keeps no state (the VMs are torn down after the build), so the live tests would need to produce some build output that gets uploaded to GitHub pages perhaps? Anyway, feel free to experiment with this.

PR #62, opened 2015-12-21
Allow different API Base for Navitia Providers

grote:
This is especially useful if you want to run a Navitia instance on your
own server.

### DIFF
--- a/enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
+++ b/enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
@@ -77,7 +77,8 @@ public abstract class AbstractNavitiaProvider extends AbstractNetworkProvider
 {
 	protected final static String SERVER_PRODUCT = "navitia";
 	protected final static String SERVER_VERSION = "v1";
-	protected static final String API_BASE = "http://api.navitia.io/" + SERVER_VERSION + "/";
+
+	protected String apiBase = "https://api.navitia.io/" + SERVER_VERSION + "/";
 
 	private enum PlaceType
 	{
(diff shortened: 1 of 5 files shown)


enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
```diff
@@ -77,7 +77,7 @@
 {
 	protected final static String SERVER_PRODUCT = "navitia";
 	protected final static String SERVER_VERSION = "v1";
-	protected static final String API_BASE = "http://api.navitia.io/" + SERVER_VERSION + "/";
+	protected static String API_BASE = "http://api.navitia.io/" + SERVER_VERSION + "/";
```
schildbach: Can you make this a proper instance variable?

enabler/test/de/schildbach/pte/live/ItalyProviderLiveTest.java
```diff
@@ -30,7 +30,7 @@
 {
 	public ItalyProviderLiveTest()
 	{
-		super(new ItalyProvider(secretProperty("navitia.authorization")));
+		super(new ItalyProvider("https://api.navitia.io/v1/", secretProperty("navitia.authorization")));
```
schildbach: This is already the default, no? Except for the s in https. If Navitia supports https, we should probably change the default.

grote:
Thanks for the review! I updated this PR.

schildbach:
I merged this with small changes. I left the HTTPS change for a future commit, actually I remembered we tried this a year ago with the ParisProvider and had some compatibility issues back then.

Thanks for your contribution!

PR #65, opened 2016-01-20
Add missing regional (TER) and long-distance (Intercités) trains support

kalon33:
This commit should fix missing regional and long-distance trains support around Paris and in South-West of France.

### DIFF
--- a/enabler/src/de/schildbach/pte/FrenchSouthWestProvider.java
+++ b/enabler/src/de/schildbach/pte/FrenchSouthWestProvider.java
@@ -53,6 +53,11 @@ public class FrenchSouthWestProvider extends AbstractNavitiaProvider
 	{
 		switch (product)
 		{
+			case REGIONAL_TRAIN:
+			{
+				// TER + Intercités
+				return new Style(Style.parseColor(color), computeForegroundColor(color));
+			}
 			case TRAM:
 			{
 				// Tram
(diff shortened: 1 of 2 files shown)


schildbach:
Merged, thanks!

schildbach:
Sorry, un-merged. It doesn't compile.

grote:
If I had to guess, I'd say that
```java
return new super.getLineStyle;
```
is not proper Java. You would need to do something like:
```java
return getLineStyle(product, code, color);
```

kalon33:
Sorry, I'm only starting with Java and pretty tired today... I followed your advice, should be good now.

kalon33:
@grote @schildbach Any updates on this? My last commit should fix the previous one.

kalon33:
@schildbach is this now possible to merge this?

schildbach:
Can you squash your two commits into one? The second is just a bugfix to the first, so we don't need them on master separately.

kalon33:
@schildbach done :)

schildbach:
It still doesn't compile. Please test your code.

I plan to use Travis to make sure the code compiles, but due to [a bug in Travis](https://github.com/travis-ci/travis-ci/issues/5227) I couldn't activate the change already.

kalon33:
`return getLineStyle(product, code, color);` probably contains an error, but as I'm only starting in Java and don't know whole public-transport-enabler code, I don't know how to correct it. An idea?

Thanks for your help.

schildbach:
No, the enum name is wrong. But the real question is: Did you ever compile this code?

grote:
@kalon33 you might want to get an IDE like AndroidStudio which can be helpful for beginners. Otherwise run `gradle build` from the command line after making code changes to see if they compile. You can run the tests with `gradle -Dtest.single=ParisProviderLive test` for example after commenting out the exclude at the end of `enabler/build.gradle`.

kalon33:
The original problem should be fixed now, and it properly builds on my machine.

schildbach:
Merged, thanks.

PR #69, opened 2016-02-06
Navitia: Fix departures query when using equiv=false

grote:
When querying departures from a `stop_area` with `equiv=false`
the query URI was wrong and returned no results.
This was especially problematic, because the suggestLocation query
returns only stop_areas at the moment.

### DIFF
--- a/enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
+++ b/enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
@@ -862,24 +862,35 @@ public abstract class AbstractNavitiaProvider extends AbstractNetworkProvider
 			// to stop_point and query departures.
 			final StringBuilder queryUri = new StringBuilder();
 			queryUri.append(uri());
-			if (equivs)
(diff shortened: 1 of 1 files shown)


schildbach:
Merged, thanks.

enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
```diff
@@ -862,24 +862,35 @@ public QueryDeparturesResult queryDepartures(final String stationId, final @Null
 			// to stop_point and query departures.
 			final StringBuilder queryUri = new StringBuilder();
 			queryUri.append(uri());
-			if (equivs)
+			final String header = stationId.substring(0, stationId.indexOf(":"));
```
schildbach: Hmmm, looks like not all Navitia IDs are separated by ":". For example the IDs used in the live tests of the OntarioProvider.

grote: That is very possible. There's a bunch of other problems with this provider as well that I am currently looking into.
Please note that I didn't add this line in this PR, but merely moved it.

PR #74, opened 2016-02-28
avoid invalid color code like "#"

stragu:
Sorry if this is not well designed as I am a newcomer, but it seemed to me that the code does something it shouldn't: the `color` value is returned even though the `IllegalArgumentException` has been returned (for example, when colorString has a value of just "#", which results in errors. (See for example https://github.com/stragu/public-transport-enabler/commit/da47f2f6a66c7ef7ec04812b1b4936bf24a1bddc#commitcomment-16374155, or https://github.com/grote/Transportr/blob/master/src/de/schildbach/pte/NzProvider.java#L49 for a check against that.)

### DIFF
--- a/enabler/src/de/schildbach/pte/dto/Style.java
+++ b/enabler/src/de/schildbach/pte/dto/Style.java
@@ -120,12 +120,16 @@ public class Style implements Serializable
 			{
 				// Set the alpha value
 				color |= 0x00000000ff000000;
+				return (int) color;
 			}
-			else if (colorString.length() != 9)
+			else if (colorString.length() == 9)
+			{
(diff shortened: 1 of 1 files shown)


schildbach:
If in doubt, it sometimes helps to write a unit test for specific cases. I just did this for the "#" case and it threw an exception. So I could not reproduce your issue.

But I took the opportunity to rewrite the code a bit (it was copy and pasted a long time ago) and added unit tests for some more cases too: See b1e89921a650b4da6a753cf58cc54605a9d1a421

stragu:
Great, thank you for looking into it! My test does pass now. Closing this useless PR.

PR #77, opened 2016-03-08
Take the location priority into account when suggesting locations in Navitia-based providers

grote:
Navitia returns a quality/priority for locations it suggests. This was ignored until now.


enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
```diff
@@ -993,10 +993,11 @@ public SuggestLocationsResult suggestLocations(final CharSequence constraint) th
 				for (int i = 0; i < places.length(); ++i)
 				{
 					final JSONObject place = places.getJSONObject(i);
+					final int priority = place.optInt("quality", 0);
```
schildbach: Just to make sure: Higher quality means more relevant, right?

grote: Even though it is not documented, it seems to mean exactly that.

schildbach:
Merged.

PR #78, opened 2016-03-08
Less network requests and no NPE when querying Navitia departures

grote:
The AbstractNavitiaProvider was making one network request per departure
it found, just to find the `Product`. This information is mostly
available in the second request it makes, so this should be taken when
available.

Also, in rare cases no destination can be found for a line. It should
not crash in these cases. This was also fixed within this commit.

Both of these issues might be problems with Navitia and should ideally
also get reported upstream.

### DIFF
--- a/enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
+++ b/enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
@@ -647,11 +647,22 @@ public abstract class AbstractNavitiaProvider extends AbstractNetworkProvider
 		try
 		{
 			final String lineId = line.getString("id");
-			final Product cachedProduct = lineProductCache.get(lineId);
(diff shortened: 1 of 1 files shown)


schildbach:
Pinging @aelkhour , I remember we talked about this "recursive fetching" of data. Did the Navitia API improve by then?

grote:
Navitia actively works to improve their service all the time and is responsive to issue reports. I reported the problem about departures missing the information we need in CanalTp/navitia#1386.

However, I observed that in some rare cases, the `physical_modes` information is missing from the `/routes` query which is what I am using at the moment. So @aelkhour's `recursive fetching` is still needed and can't be removed. I hope that this won't be necessary anymore once CanalTp/navitia#1386 gets fixed.

schildbach:
Ok, tested this a bit and merged. Thanks!

PR #79, opened 2016-03-09
Navitia: Handle empty or non-present color attributes, …

schildbach:
…use the product default colors in these cases.

@grote @aelkhour


grote:
Looks good to me!

PR #80, opened 2016-03-11
Add a name to Line and use it with Navitia providers

grote:
Also add network to Line in Navitia providers

A URL for a line does not seem to be exposed via the Navitia API.

Due to a bug uncovered by adding the network to the line,
the code to retrieve departures had to be simplified.
It now does not do an extra network request to get the line destination.

### DIFF
--- a/enabler/src/de/schildbach/pte/dto/Line.java
+++ b/enabler/src/de/schildbach/pte/dto/Line.java
@@ -43,6 +43,7 @@ public final class Line implements Serializable, Comparable<Line>
 	public final @Nullable String network;
 	public final @Nullable Product product;
 	public final @Nullable String label;
+	public final @Nullable String name;
 	public final @Nullable Style style;
 	public final @Nullable Set<Attr> attrs;
 	public final @Nullable String message;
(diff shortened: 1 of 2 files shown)


schildbach:
Should the name added to the toString()?

enabler/src/de/schildbach/pte/dto/Line.java
```diff
@@ -54,31 +55,58 @@
 
 	public Line(final String id, final String network, final Product product, final String label)
 	{
-		this(id, network, product, label, null, null, null);
+		this(id, network, product, label, null, null, null, null);
+	}
+
+	public Line(final String id, final String network, final Product product, final String label, final String name)
```
schildbach: I think we only need to add the 1 constructor you're using, and the new version with all parameters. Others can be added as needed.

grote:
I updated the PR. Unfortunately, it now got bigger than originally intended. Let me know what you think.

grote:
Please let me know once you had a chance to look at this one.

grote:
This branch now has conflicts. Want me to rebase?

schildbach:
Yes, please. I just discovered the conflict, too (-:

grote:
Alright, rebased. Was more difficult than expected, so you might want to review carefully.

schildbach:
Thanks, merged.

Just a minor note: The textcolor change could have gone on a separate commit (if not PR), since its entirely independent of the "add name" change. The smaller/more concise a PR is, the easier it is to review and merge.

PR #81, opened 2016-03-13
Properly create Locations with name and place in Navitia providers

grote:
Allow option to fix case of location names and simplify location parsing
code.

### DIFF
--- a/enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
+++ b/enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
@@ -211,7 +211,7 @@ public abstract class AbstractNavitiaProvider extends AbstractNetworkProvider
 			}
 			case POI:
 			{
-				return LocationType.ANY;
+				return LocationType.POI;
 			}
 			default:
 				throw new IllegalArgumentException("Unhandled place type: " + placeType);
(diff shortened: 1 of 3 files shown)


grote:
I added another related commit that includes `POI`s in the `suggestLocations` result.

You better merge these PRs or I keep adding stuff ;)

grote:
Are there any issues that block merging this PR?

schildbach:
No worries, I just took a little break (-:

I merged your first commit. About the second commit, I wonder if those additional suggestions will work with trip queries. Are there already unit tests for such trip queries?

In future I think it'd be better to separate such commits into two PRs; I see to connection between them.

grote:
>No worries, I just took a little break (-:

Good to have you back! :)

>I merged your first commit.

Thanks!

> About the second commit, I wonder if those
>additional suggestions will work with trip queries. Are there already
>unit tests for such trip queries?

No, I can make them. Will you if they are there?

>In future I think it'd be better to separate such commits into two PRs;
>I see to connection between them.

Will do.

schildbach:
Yes, I'm fine with the change. I wonder if we should add an `EnumSet<LocationType> types` parameter to `suggestLocations()`, similar to `queryNearbyLocations()`.

Clarification to my previous comment: I meant "live test"…

grote:
I'd be fine adding a `EnumSet<LocationType> types` parameter to `suggestLocations()`. How will you handle that in the other providers?

For the administrative regions, should we introduce a new location type or use either `ADDRESS` or `POI`. Currently, I think I am using `POI`.

grote:
So I added a live test. Is it sufficient?

schildbach:
Yes, that looks good. Thanks, merged.

grote:
Wohoo! Thank you! :)

PR #85, opened 2016-04-14
Add gradle wrapper

JonasGroeger:
Allows to get started easier, i.e.

```
git clone ...
cd ...
./gradlew -Dtest.single=VgnProviderLive
```


grote:
I personally don't like the gradle wrapper since it loads executables from the internet in an insecure fashion to execute on people's computer.

schildbach:
We decided against using gradle wrapper. It's easy to install gradle if you don't have it, e.g. on Debian/Ubuntu `sudo apt-get install gradle`.

PR #88, opened 2016-05-22
Navitia: Support cablecar and other unknown products in Paris

grote:


schildbach:
I merged the cablecar change, but not the default change. I know throwing exceptions is a bit rude, but at least as a developer you notice something is wrong.

Even if we do want the default change, I think it should go on a different PR.

grote:
Thanks for merging so quickly! :)

> I know throwing exceptions is a bit rude, but at least as a developer you notice something is wrong.

Sure, but I as a developer don't know all products that are available in the region. I just found the cablecar problem by accident. Most people just don't get a connection. The FR-SW has the same problem.

I think since this is "just" about line styles, throwing an exception is a bit much and a default style should be shown instead. A developer with knowledge in the region might still be able to tell that the style is wrong.

schildbach:
Fair enough. I just added your change as a second commit.

PR #89, opened 2016-05-23
Navitia: Find products for nearby stations

grote:
If available, the locations found in the nearby stations query, will now have products.


schildbach:
Great! Merged.

PR #96, opened 2016-05-31
Navitia: More efficient product parsing

grote:
According to CanalTP/navitia#1386, Navitia has finally added the
product to departure query results, so all information we need can now
be retrieved by one network query only.

The AbstractNavitiaProvider was adapted accordingly and all legacy code
including the Product Cache was removed.

### DIFF
--- a/enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
+++ b/enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
@@ -669,15 +669,18 @@ public abstract class AbstractNavitiaProvider extends AbstractNetworkProvider
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
```diff
@@ -669,15 +669,18 @@ private void parseQueryTripsResult(final JSONObject head, final Location from, f
 		}
 	}
 
-	private Line parseLine(final JSONObject jsonLine) throws IOException
+	private Line parseLine(final JSONObject jsonRoute) throws IOException
```
schildbach: Shouldn't this method be renamed to parseRoute?

schildbach:
That's great to hear! Thanks for the patch, merged.

Issue #84, opened 2016-04-13
Add New York, USA provider


anarcat:
Just a feature request for now.. It seems that Navitia has the datasets for NY that we can use:

https://navitia.opendatasoft.com/explore/dataset/us-ny/

I am not sure how to make the necessary code happen here, but I'd love to see that happen, so just throwing this out there until I can find time / resources to actually learn how to implement this!

schildbach:
You could use ParisProvider or ItalyProvider (or any other existing subclass of AbstractNavitiaProvider) as a template.

schildbach:
Thanks a lot for your report. Unfortunately, due to the current bad shape of Navitia support in public-transport enabler, time constraints on our side and due to the centralized nature of navitia.io I have decided to drop support for Navitia. Thus, I'm going to close this report. Sorry for the inconvenience and thanks again.

PR #97, opened 2016-06-05
Add New York, USA provider

chimo:
Ref. #84, grote/Transportr#180

### DIFF
--- a/enabler/src/de/schildbach/pte/NewyorkProvider.java
+++ b/enabler/src/de/schildbach/pte/NewyorkProvider.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


enabler/src/de/schildbach/pte/NewyorkProvider.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * @author Stephane Berube
+ */
+public class NewyorkProvider extends AbstractNavitiaProvider
+{
+	private static String API_REGION = "us-ny";
+
+	public NewyorkProvider(final String apiBase, final String authorization)
+	{
+		super(NetworkId.QUEBEC, apiBase, authorization);
```
grote: It probably should not use `QUEBEC` here.

enabler/src/de/schildbach/pte/NewyorkProvider.java
```diff
@@ -0,0 +1,46 @@
...
+
+		setTimeZone("America/Montreal");
```
grote: This should probably say `New_York` as well.

chimo:
@grote thanks for the comments. Guess I was tired when I did this :|

leandroferreira:
@grote @chimo Is there any additional work required on this PR before it can be merged?

grote:
Not that I know of. You can practically test this in Transportr already and let us know how well it works. @schildbach is the only one that can merge PRs here.

PR #109, opened 2016-09-21
Switch to Java 7 and use OkHttp

schildbach:
Notes regarding Android compatibility:
- The 'try-with-resources' language feature needs API level 19, so don't use it.
  All other language features can be used back to API level 8.
- The JDK 1.7 API still isn't fully implemented, so better stick with 1.6 for now.

Migrate all HTTP calls to use OkHttp rather than URLConnection.


schildbach:
Merged.

Issue #108, opened 2016-09-15
Support France-North-West


patkan:
Hi, I would like to see support for France-North-West (especially Rennes).
There is Navitia-data available as "fr-nw". I have forked this project and started to add fr-nw in https://github.com/patkan/public-transport-enabler/tree/add-fr-nw-provider.
However I am currently stuck at testing, since I applied for a Navitia-account but haven't been activated yet.
If someone has the time and a Navitia-account I would be happy if he or she could verify my live-tests.
I will create a Pull Request as soon as I have managed to test this.

grote:
If you still didn't get an API token, try again. They sometimes seem to loose the requests...

patkan:
Thank you that's good to know. I am for a few days on the road and will try afterwards.

schildbach:
Fixed in #110.

PR #110, opened 2016-09-23
add basic support for fr-nw-region

patkan:
As requested in #108 I tried to add fr-nw from navitia into pte.
I have run the live tests locally and they build succesfully. Please let me know if more tests are needed or anything else has to be fixed.


schildbach:
Merged, thanks!

Btw. you don't need to open a new PR if you refine an existing one. Just force-push your branch over the old one, and GitHub will update accordingly.

PR #125, opened 2016-12-09
Quick fix for missing on-demand section type

### LINKED ISSUE #123: Navitia: Unhandled place type: on_demand_transport
I have found a problem in the FranceNorthWest-Provider.
Search a route from "La Villeneuve, Missillac" to "IUT, Rennes". It will yield in Transportr a toast telling me `IllegalArgumentException: Unhandled place type: on_demand_transport`.

After looking into the code I think the problem is that this is not implemented in the Navitia-provider. This type of transport is described in the [Navitia-API-documentation](http://doc.navitia.io/#on-demand-transportation).

I don't have the time at the moment to fix this, but for later or someone else: is this a problem that should be fixed in the NavitiaProvider or in the implementation which is in this case FranceNorthWestProvider? (Or am I misunderstanding something?)

grote:
Fixes #123

### DIFF
--- a/enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
+++ b/enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
@@ -400,7 +400,7 @@ public abstract class AbstractNavitiaProvider extends AbstractNetworkProvider {
         }
     }
 
-    private Line parseLineFromSection(final JSONObject section) throws IOException {
+    private Line parseLineFromSection(final JSONObject section, final SectionType type) throws IOException {
         try {
             final JSONArray links = section.getJSONArray("links");
             String lineId = null;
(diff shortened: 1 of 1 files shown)


grote:
@patkan check if that helps.

patkan:
Thanks @grote, this fixes the problem. Transportr can now list the on-demand-products. With this version I cannot reproduce the documented error.
In Transportr we should maybe show a warning that this is on-demand, but this does not belong here.

schildbach:
Looks good, but if this is a quickfix what would be the thorough fix?

grote:
Sorry for the misleading terminology. Quick fix in this case just means that I did it without researching how the API response is supposed to look like and without testing it with a variety of providers.

schildbach:
Thanks, merged.

Issue #107, opened 2016-09-02
NPE in France North East Provider


grote:
```
09-02 12:24:07.911 W/System.err: java.lang.NullPointerException
09-02 12:24:07.911 W/System.err:     at com.google.common.base.Preconditions.checkNotNull(Preconditions.java:210)
09-02 12:24:07.911 W/System.err:     at de.schildbach.pte.dto.Style.parseColor(Style.java:116)
09-02 12:24:07.911 W/System.err:     at de.schildbach.pte.FranceNorthEastProvider.getLineStyle(FranceNorthEastProvider.java:66)
09-02 12:24:07.911 W/System.err:     at de.schildbach.pte.AbstractNavitiaProvider.parseLineFromSection(AbstractNavitiaProvider.java:494)
09-02 12:24:07.911 W/System.err:     at de.schildbach.pte.AbstractNavitiaProvider.parseLeg(AbstractNavitiaProvider.java:565)
09-02 12:24:07.911 W/System.err:     at de.schildbach.pte.AbstractNavitiaProvider.parseQueryTripsResult(AbstractNavitiaProvider.java:658)
09-02 12:24:07.911 W/System.err:     at de.schildbach.pte.AbstractNavitiaProvider.queryTrips(AbstractNavitiaProvider.java:1144)
```

grote:
@stragu that's your provider, right? You can reproduce by searching for trips from "Metz, GARE" to "Stasbourg, Kibitzenau".

frikilax:
I also have the issue, again with France north east provider 
I just remembered this is why I don't use Transportr, I just can't...
(I can't give much details right now, but I will later)

PR #127, opened 2016-12-18
fixes NPE: sets default color to grey instead of null if not defined in the configuration

ghost:
Fix to #107, with the help of @corentin38
NPE seemed to occur also on France South-West and South-East regions too.

### DIFF
--- a/enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
+++ b/enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
@@ -421,7 +421,7 @@ public abstract class AbstractNavitiaProvider extends AbstractNetworkProvider {
             final String code = displayInfo.getString("code");
             final String color = Strings.emptyToNull(displayInfo.getString("color"));
             final String name = Strings.emptyToNull(displayInfo.optString("headsign"));
-            final Style lineStyle = getLineStyle(product, code, color != null ? "#" + color : null);
+            final Style lineStyle = getLineStyle(product, code, color != null ? "#" + color : "#8a000000");
 
             return new Line(lineId, network, product, code, name, lineStyle);
         } catch (final JSONException jsonExc) {


patkan:
I have tested this PR by pulling in the changes locally into pte in Transportr and can confirm that it fixes #107.

schildbach:
Ah sorry I should have commented this earlier. I'm not convinced that defaults should be defined in PTE. Imho, if the interface doesn't have a color it should say so (by using the value 'null') and the app should decide on a default color.

What do you think?

grote:
I wouldn't mind this being set by PTE, maybe also on a per-provider basis as the default color probably varies from provider to provider.

The problem with letting the individual providers take care of it is that they don't handle the `@Nullable` color properly and cause NPEs. We had that in the past a lot and it is still happening.

Maybe a protected `getDefaultColor()` method that providers can override if they like?

schildbach:
Yes, you generally have to handle null values. There are lots of values that can be null in the PTE results. Maybe we need to add more `@Nullable` annotations?

grote:
I would be great if all return values that can be `null` would be annotated `@Nullable`, same for parameters.

But I see this discussion more about where/how/if default colors should be handled.

if `getLineStyle()` would not be overwritten in the [FranceNorthEastProvider](https://github.com/schildbach/public-transport-enabler/blob/master/enabler/src/de/schildbach/pte/FranceNorthEastProvider.java#L52) (and other Navitia providers), it would use a default color and shape for lines, right? So that is already handled/set somewhere (I haven't checked where).

We could make this more modular and use as an automatic fallback in other places when the color is `null` or just give the app total responsibility.

When viewing PTE as a versatile and easy-to-use library, it would be a nice addition to provide means to set a default color that is automatically used if no color is there, so this logic does not need to be re-implemented by every single program using PTE.

schildbach:
I'll try to go through the API and add `@Nullable` where appropriate. I just noticed that colors are ints, and in fact the value `0` is used if there is no color. This is congruent with the Android API, but of course PTE is not only used by Android developers. We could switch all colors to `Integer`, but I'm not sure if it's worth it.

There are some style defaults, but only for products that have a style default in the real world. For example, in Germany S-Bahn has this dark green rounded style. In these cases, defaults are ok. But I think PTE generally – not only for styles – should not invent values. Imho the handling of missing data is up to the app using PTE.

PR #128, opened 2016-12-30
Use https for several providers

full-duplex:
Some providers (MERSEY, MVV, NASA, NS, NVBW, SNCB and VRR) still use an insecure url while the tests still pass when using a secure url. This pull request changes them to use https instead.

### DIFF
--- a/enabler/src/de/schildbach/pte/AbstractHafasProvider.java
+++ b/enabler/src/de/schildbach/pte/AbstractHafasProvider.java
@@ -3036,7 +3036,8 @@ public abstract class AbstractHafasProvider extends AbstractNetworkProvider {
         uri.append('y');
         uri.append("?performLocating=2&tpl=stop2json");
         uri.append("&look_stopclass=").append(allProductsInt());
-        uri.append("&look_nv=get_stopweight|yes");
+        // The | character needs to be urlencoded otherwise SNCB fails
+        uri.append("&look_nv=get_stopweight%7Cyes");
         // get_shortjson|yes
         // get_lines|yes
         // combinemode|2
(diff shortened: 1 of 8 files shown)


schildbach:
Did you test all these URLs?

full-duplex:
Yes, these are all the providers that were still using an http url, had at least some passing tests and the same amount of passing live tests when switched over to https.

One exception is SNCB where one test failed with the secure url, but this is fixed in the second commit.

schildbach:
Great to hear. Seems like the companies feel a real push towards HTTPS. I checked these providers only a couple of weeks ago.

I'll merge this at the next occasion. The URL encoding bug I can't reproduce and also the '|' sign doesn't need to be encoded.

full-duplex:
Are you sure? This is the url that the test uses and I can reproduce it in mutliple browsers and curl, confirming it's a server side issue.

```
https://www.belgianrail.be/jp/sncb-nmbs-routeplanner/query.exe/nny?performLocating=2&tpl=stop2json&look_stopclass=65535&look_nv=get_stopweight|yes&look_x=3407118&look_y=50748017&look_maxno=200&look_maxdist=5000
```

In fact, any url that has a pipe in it fails combined with https:

```
$ curl "https://www.belgianrail.be/|" 
curl: (56) GnuTLS recv error (-54): Error in the pull function.

$ curl "https://www.belgianrail.be/%7C"
<!DOCTYPE html>
...
```

schildbach:
Just to record the state, I merged your first commit a while ago. The second one I still need to re-review. If you could rebase this PR it would be great.

full-duplex:
While rebasing the fix, I found out that the SNCBProvider is broken. It looks like they removed the API that it relies on. This PR can be closed and the SNCBProvider should probably be removed.

schildbach:
I switched SncbProvider back to HTTP, that fixes the problem.

PR #129, opened 2017-01-09
NegentweeProvider: Implementation of 9292.nl provider

full-duplex:
This adds a provider for the Dutch 9292.nl, which includes all public transport in The Netherlands.

Their API is non-standard so it's implemented from scratch and includes tests.

### DIFF
--- a/enabler/src/de/schildbach/pte/NegentweeProvider.java
+++ b/enabler/src/de/schildbach/pte/NegentweeProvider.java
@@ -1,0 +1,1019 @@
+/*
+ * Copyright 2010-2015 the original author or authors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package de.schildbach.pte;
+
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.io.IOException;
+import java.io.Serializable;
+import java.text.ParseException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Currency;
+import java.util.Date;
+import java.util.EnumSet;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.text.SimpleDateFormat;
+
+import javax.annotation.Nullable;
+
+import de.schildbach.pte.dto.Departure;
+import de.schildbach.pte.dto.Fare;
+import de.schildbach.pte.dto.Line;
+import de.schildbach.pte.dto.LineDestination;
+import de.schildbach.pte.dto.Location;
+import de.schildbach.pte.dto.LocationType;
+import de.schildbach.pte.dto.NearbyLocationsResult;
+import de.schildbach.pte.dto.Point;
+import de.schildbach.pte.dto.Position;
+import de.schildbach.pte.dto.Product;
+import de.schildbach.pte.dto.QueryDeparturesResult;
+import de.schildbach.pte.dto.QueryTripsContext;
+import de.schildbach.pte.dto.QueryTripsResult;
+import de.schildbach.pte.dto.ResultHeader;
+import de.schildbach.pte.dto.StationDepartures;
+import de.schildbach.pte.dto.Stop;
+import de.schildbach.pte.dto.SuggestLocationsResult;
+
+import de.schildbach.pte.dto.SuggestedLocation;
+import de.schildbach.pte.dto.Trip;
+import de.schildbach.pte.exception.InternalErrorException;
+import de.schildbach.pte.exception.NotFoundException;
+import de.schildbach.pte.util.ParserUtils;
+import de.schildbach.pte.util.WordUtils;
+import okhttp3.HttpUrl;
+
+/**
+ * @author full-duplex
+ */
+public class NegentweeProvider extends AbstractNetworkProvider {
+
+    private static final String API_BASE = "https://api.9292.nl/0.1/";
+    private static final String SERVER_PRODUCT = "negentwee";
+
+    private static final Language DEFAULT_API_LANG = Language.NL_NL;
+    private static final int DEFAULT_MAX_LOCATIONS = 50;
+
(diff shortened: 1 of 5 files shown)


schildbach:
Thanks! This generally looks good to me. I'll add a couple of code comments.

enabler/src/de/schildbach/pte/NegentweeProvider.java
```diff
@@ -0,0 +1,986 @@
...
+    private final Language language;
+    private final ResultHeader resultHeader;
+
+    public enum Language {
+        NL_NL("nl-NL"),
+        EN_GB("en-GB");
+
+        private final String lang;
+
+        Language(String lang) {
```
schildbach: Why no visibility?

enabler/src/de/schildbach/pte/NegentweeProvider.java
```diff
@@ -0,0 +1,986 @@
...
+
+        public String toString() {
+            return name().toLowerCase();
+        }
+    }
+
+    public static class QueryParameter implements Serializable {
+        public String name, value;
+
+        QueryParameter(String name, String value){
```
schildbach: Same here.

enabler/src/de/schildbach/pte/NegentweeProvider.java
```diff
@@ -0,0 +1,986 @@
...
+        throw new RuntimeException("Cannot find station with name " + locationName);
+    }
+
+    private List<Location> queryLocationsByName(String locationName, EnumSet<LocationType> types) throws IOException {
+        ArrayList<QueryParameter> queryParameters = new ArrayList<>();
+        queryParameters.add(new QueryParameter("q", locationName));
+
+        if (!types.contains(LocationType.ANY) && types.size() > 0) {
+            StringBuilder typeValue = new StringBuilder();
+            for(LocationType type : types) {
```
schildbach: Space missing behind "for" (repeated multiple times)

enabler/src/de/schildbach/pte/NegentweeProvider.java
```diff
@@ -0,0 +1,986 @@
...
+                                legMessage.append(legDisturbance.getString("effect"));
+                                legMessage.append(" ");
+                                legMessage.append(legDisturbance.getString("measure"));
+                            }
+                        }
+                    }
+
+                    foundLegs.add(new Trip.Public(
+                            new Line(
+                                    leg.getString("service"), // final String id
```
schildbach: I'm not sure if these comments are useful. Your IDE should be able to tell the parameter names.

full-duplex: My Android Studio was a bit stubborn in showing them so I added them during development. But you're right, I've removed them.

enabler/src/de/schildbach/pte/NegentweeProvider.java
full-duplex: Fix this for all `if` and `for` statements that had this problem.

enabler/src/de/schildbach/pte/NegentweeProvider.java
full-duplex: This class is an implementation detail only used internally that I made public so I could use it in a test during development. I made the whole class private instead (if you want to reuse it somewhere and make it public, it should probably be moved to another file).

enabler/src/de/schildbach/pte/NegentweeProvider.java
full-duplex: This constructor can't be public (I get an compilation error). Probably because it's an `Enum`?

full-duplex:
Thanks for your feedback, I fixed the problems you've raised. Let me know if you are ready to merge it like this.

schildbach:
Thanks! I have one more question: What does the name "Negentwee" means and how is it related to 9292.nl?

full-duplex:
Negentwee means Ninetwo (Negen means nine, twee means two). I've encountered them used that name internally, so I copied it because naming a class or Enum option 9292 can be problematic.

schildbach:
I'm a bit sceptical about a made-up name here, but you're right we can't name the class 9292Provider.java. What about NineTwoNineTwoProvider.java?

(Of course this is not a blocker for merging.)

full-duplex:
It's a bit long, which is one of the reasons why I went with their internal name. If you prefer, I can change it though.

schildbach:
No ok it's fine. I misread your first answer.

full-duplex:
Ok, great.

schildbach:
I'm just reviewing your PR more thoroughly. Very important: Can you add the license header for your added source files? Also, you could put yourself into the authors similar to the other providers.

I've also seen a few missing `@Override` at the toString()'s.

schildbach:
The unit tests run great!

schildbach:
One more problem: Your code uses Java8. We're not there yet, can you step back to Java7 for now?

schildbach:
Ah sorry, forget my last last comment. Was thinking type inference is Java8, but it's Java7.

enabler/src/de/schildbach/pte/NegentweeProvider.java
```diff
@@ -0,0 +1,997 @@
...
+
+        // Default query options
+        ArrayList<QueryParameter> queryParameters = new ArrayList<>();
+        queryParameters.add(new QueryParameter(
+                "latlong",
+                location.getLatAsDouble() + "," + location.getLonAsDouble()
+        ));
+        queryParameters.add(new QueryParameter(
+                "rows",
+                String.valueOf((maxLocations <= 0) ? DEFAULT_MAX_LOCATIONS : maxLocations))
```
schildbach: The server seems to fail with maxLocations > 100, so we should silently cap this parameter to 100.

full-duplex: Good catch! Fixed this.

full-duplex:
Thanks for the thorough review. I've added the copyright statements and `@Override` annotations and fixed the `maxLocations` bug.

schildbach:
Merged (with minor changes)! Thanks a lot for your effort!

While reviewing again, I noticed your QueryParameter abstraction. I think HttpUrl has a nice API for parameters already – is there a reason you didn't use that directly?

grote:
Great to see this merged. I gave it a quick test with Transportr and noticed that the `Line.label` is not set as with other providers. Here, it uses the name of the product such as "Intercity", "Sprinter", but not the number like "IC234".

full-duplex:
@schildbach Great! The `QueryParameter` abstraction is mostly because of the way the code evolved and because I overlooked the parameter API until after I made the PR. It's probably not too difficult to remove that class if you want. 

@grote The omission of service numbers for trains in the `Line.label` is intentional. Train service numbers have no real meaning (they aren't mentioned anywhere and the official 9292 website and app omit them as well) and names such as "Intercity Direct 12345" are unnecessarily long without any benefit to the user. It's not possible to abbreviate that either, because those names can be different depending on the language and the provider that runs the line. This applies only to trains. Any other transportation method has their service number set in the label.

grote:
> This applies only to trains. Any other transportation method has their service number set in the label.

OK fine with me. Thanks for explaining. Seems I only got trains when testing.

PR #130, opened 2017-01-18
fix NPE in AbstractNavitiaProvider

patkan:
4f7bd832e68527a343710d69d11b179cd49116a6 introduced a NullPointerException in AbstractNavitiaProvider. The call to `Context()` in line 970 calls `prevQueryUrl.toString()` and `nextQueryUrl.toString()`.
The variables are however `null` in the case that the search returned only one result. The easiest way to reproduce this is searching for two very nearby locations. The result will then be one walk-route.
This commit fixes the NPE by restoring the old behavior and has been tested in Transportr.


schildbach:
Thanks, merged.

PR #132, opened 2017-01-20
Don't swallow trips in Navitia providers

grote:
Use the `min_nb_journeys` parameter rather than the `count` parameter when requesting trips.

The `count` parameter causes Navitia to remove trips arbitrarily from the result which can look like a bug to the user when she knows a trip should be there, but it isn't.


schildbach:
LGTM, but I didn't test it yet.

grote:
ping ;)

schildbach:
Merged! Sorry for the delay.

PR #140, opened 2017-01-31
Made SuggestedLocations public

### LINKED ISSUE #138: Expose SuggestedLocation to library consumers
Would you accept a PR that makes the `SuggestedLocation`s in `SuggestedLocationResult` public?

The use case is advanced sorting of auto-completion results that should take the priority value returned by the APIs into account.

grote:
so their priority can be used for custom sorting

Closes #138


schildbach:
Merged.

PR #146, opened 2017-02-21
Add support for Navitia Costa Rica provider.

jamescr:
The Navitia server running is a locally installed version. This provider shares data from the Costa Rica Inter urban Train Network (Incofer Tren Urbano) and hopefully will be added to Transport soon. :-)

### DIFF
--- a/enabler/test/de/schildbach/pte/live/CostaRicaProviderLiveTest.java
+++ b/enabler/test/de/schildbach/pte/live/CostaRicaProviderLiveTest.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2010-2015 the original author or authors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package de.schildbach.pte.live;
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Test;
+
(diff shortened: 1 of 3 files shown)


enabler/test/de/schildbach/pte/live/CostaRicaProviderLiveTest.java
```diff
@@ -0,0 +1,89 @@
...
+import de.schildbach.pte.dto.NearbyLocationsResult;
+import de.schildbach.pte.dto.QueryDeparturesResult;
+import de.schildbach.pte.dto.SuggestLocationsResult;
+
+/**
+ * @author Jaime Gutiérrez Alfaro
+ */
+public class CostaRicaProviderLiveTest extends AbstractNavitiaProviderLiveTest {
+    public CostaRicaProviderLiveTest() {
+        super(new CostaRicaProvider(secretProperty("navitia.authorization")));
```
grote: If your server requires authorization (does it?), it is probably different from the one of the official navitia server, right? So in that case this should probably use a different secret variable here.

jamescr: This server does not require authorization and I don't think will be massively used. I think I could change the code, from "navitia.authorization" to null. 

Also given that the server is administrated by a public university, no one disagree to keep it without authorization.  Do you recommend to add the authorization?

grote: I personally don't like API secrets embedded in the source code. They are not secure anyway and don't work well with reproducible free software.

However, my point here was that this code above will attempt to use the secret property from the official navitia server. So it should either be removed or changed. Since your server doesn't require authorization, I'd just remove it. The live tests continue to work then, right?

jamescr: I have removed it. Thanks. 

The tests continue working without the secrets.

jamescr:
As testing feedback, I used this PR code to generate a public-transport-enabler library, that later i used to compile Transportr (v1.1.8).

grote:
@jamescr do you have any students or other people using this regularly who can attest to the quality, so it can be included in PTE? If I understand correctly, such third-party attestations will help @schildbach whether to merge or not.

Sufrostico:
I tested this apk version, And it's working pretty good

schildbach:

Issue #147, opened 2017-02-25
NPE in SuggestLocationsResult


sbiermann:
Hi,
when a EFA Provider is not correctly answering when searching for a location a NPE happens in SuggestLocationsResult when getLocations() is called.

Currently the VRN Provider (EFA-Provider) doesn't work correct, so the Query: 
http://fahrplanauskunft.vrn.de/vrn/XML_STOPFINDER_REQUEST?outputFormat=JSON&language=de&stateless=1&coordOutputFormat=WGS84&locationServerActive=1&regionID_sf=1&type_sf=any&name_sf=Heidelberg&anyObjFilter_sf=126&reducedAnyPostcodeObjFilter_sf=64&reducedAnyTooManyObjFilter_sf=2&useHouseNumberList=true&anyMaxSizeHitList=500
returns:
{ "parameters": [ { "name":"requestID", "value":"0" }, { "name":"sessionID", "value":"0" } ], "stopFinder": { "message": [ { "name":"code", "value":"-8020" }, { "name":"error", "value":"" }, { "name":"type", "value":"" }, { "name":"module", "value":"BROKER" } ], "input": { "input":"Heidelberg" }, "points": null } }

With this answer the field suggestedLocations in SuggestLocationResult is null and getLocations() throws an NPE when it is called.

schildbach:
From the URL you posted it looks like you're using an old version. VrnProvider is using HTTPS these days.

I tried to reproduce your problem by changing one of the suggestLocations* unit tests to enter "Heidelberg". It yields the same JSON result, but ultimately parses this correctly to SERVICE_DOWN (note an error -8020 is indicated in the JSON).

Can you rebase to current master (if not done already) and then try if your problem shows up in an unit test too?

sbiermann:
I have updated to the current master (this project is used as git submodule) and now it works without any fix. Seems that there was another fix which solves the problem.

PR #148, opened 2017-02-25
adding NPE check

sbiermann:
Fix for #147


schildbach:
Even if it turns out a fix is needed, I'm almost sure making up an empty array is not the right fix. See my other comment on #147.

PR #149, opened 2017-02-25
NegentweeProvider: Skip trips that have cancellations

full-duplex:
This should fix a nasty bug where cancelled trips were returned as if they were still possible.


full-duplex:
Also, it seems that this is the best way to handle cancelled trips? There is no way to mark a trip as fatally delayed or cancelled so that an application can still show the trip?

schildbach:
Merged, thanks. Up to now, there is no "cancelled" marker in PTE. This is because Öffi is focused on the realtime data, not on the plan data.

PR #155, opened 2017-03-22
add getNumChanges

### LINKED ISSUE #151: numChanges of Trips null in AbstractHafasMobileProvider
Over at Transportr we have an issue with the returned trips from provider that are based on `AbstractHafasMobileProvider`. The member `numChanges` of the returned trips is `null`. Please see grote/Transportr#316 for reference. I tried to summarize my findings in [this post](https://github.com/grote/Transportr/issues/316#issuecomment-283808598).

May the value `numChanges` be `null`? If yes, could you maybe annotate the `Trip`-class accordingly? In this case we have to adapt the code of Transportr of course.

Is the number of changes not available in the new API or why is this value set to `null`?

patkan:
Computes the number of changes on a trip for numChanges that is null.
I am not sure how to handle the case when two Public Legs are connected by an Individual. In this case this would be also counted as one change. At least DBProvider copies also has this behavior. All in all I think I copied the behavior of the other APIs. I compared with VGN and DB.

fixes #151

### DIFF
--- a/enabler/src/de/schildbach/pte/dto/Trip.java
+++ b/enabler/src/de/schildbach/pte/dto/Trip.java
@@ -150,6 +150,34 @@ public final class Trip implements Serializable {
         return maxTime;
     }
 
+    /** Number of Changes on the trip.
+     *
+     * Returns the variable numChanges if it is not null.
+     * If numChanges is null, it tries to reconstruct the number of changes where applicable.
+     * The number of changes for a Trip consisting of only individual Legs is null.
+     *
+     * @return Number of changes on the trip
+     */
+    @Nullable
+    public Integer getNumChanges() {
+        if (numChanges == null) {
+            Integer numCount = null;
+
+            for (final Leg leg : legs){
+                if (leg instanceof Public) {
+                    if (numCount == null) {
+                        numCount = 0;
+                    } else {
+                        numCount++;
+                    }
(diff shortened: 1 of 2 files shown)


enabler/src/de/schildbach/pte/dto/Trip.java
```diff
@@ -150,6 +150,35 @@ public Date getMaxTime() {
...
+     *
+     * @return Number of changes on the trip
+     */
+    @Nullable
+    public Integer getNumChanges() {
+        if (numChanges == null) {
+            Integer numCount = null;
+
+            for (final Leg leg : legs){
+                if (!(leg instanceof Individual)) {
```
schildbach: Why not explicitly test for `leg instanceof Public` here?

patkan: I mistook `Leg` for being instatiable. So I tried to test for the case that we don't have an `Individual`. I will change this to a test for `Public`, because that is more explicit.

patkan:
I amended my last commit with the change from `Individual` to `Public`. Also I removed my author-Tag because that's just unneeded noise in a method-javadoc (and according to my style-checker does not even belong there).

schildbach:
Merged (with some reformatting and wording changes). Thanks!

Issue #159, opened 2017-04-22
Norway: Wrong Base URL in NriProvider.


snake66:
The URL used by NriProvider (http://hafas.websrv05.reiseinfo.no/bin/dev/nri/) is no longer valid. The service has been moved to another URL (http://apiprod.reiseinfo.no/bin/rest.exe/v1.1/) Just switching the Base URL should work according to Entur which has taken over responsibility for this service.

schildbach:
Tried switching over the base URL doesn't work, it requires some form of authentication. Can you link the statement from Entur?

snake66:
I asked them by email, and they gave me the new url and say to use the same auth_key as previously. Perhaps the old service worked without an auth_key and the new one don't?

They do have a [form](http://labs.rutebok.no/aapent_api.aspx) for registering as users of the API that will give you the auth key and some more information. They will also keep you posted on further changes in the API and service. (They warn that there will be further/bigger changes coming.)

Not sure how you prefer to go about these things, if each user of the library should get their own auth_keys or you want to provide them as part of the library. In any case, I'll be happy to assist in keeping this provider up to date. (I rely on the info myself.)

schildbach:
I suspect the API they are replacing is not the one public-transport-enabler is/was using. But anyway, if you register maybe you can post a link to their API description so we can have a look.

There are a couple of providers that already require some form of authorization, e.g. all Navitia providers. Have a look at AbstractNavitiaProvider, we could handle it in a similar manner. Basically, the app owner register their apps separately.

snake66:
Seems like a sensible approach, I'll register and see where that leads. An alternative may be to use the Ruter API as mentioned in #150. Seems like it does not require authorization (for now at least.)

snake66:
Their registration form didn't work. I've contacted them by email again.

BredeD:
Hi, 
Here is an example URL. You can use the combination /vs_restapi/ and authKey below.
http://apiprod.reiseinfo.no/bin/rest.exe/v1.1/vs_restapi/trip?authKey=api-test&originId=761023999&destId=761008829

I think this should work. Remove other traces of old 'vs_' and authKey.
Best regards
Brede / Entur as

schildbach:
Ok that API is currently not supported by PTE. The XSD for querying trips is available here: http://apiprod.reiseinfo.no/bin//rest.exe/vs_restapi/v1.1/xsd?name=hafasRestTrip.xsd

[...]

PR #160, opened 2017-05-19
NriProvider: Update base url.

snake66:
This is a temporary fix for #159 that allows us to use the old api until Entur makes the new API official.

This patch passes all the NriProviderLiveTest cases _except_ the short trips test. I don't know what's causing it but it fails with the following exception:

    java.lang.IllegalStateException: error C19 Communication: Read failed
            at de.schildbach.pte.AbstractHafasLegacyProvider$2.onSuccessful(AbstractHafasLegacyProvider.java:835)

I just wanted to submit the pr now in case it's something obvious that @schildbach or anybody else can point out to me. Otherwise I'll keep debugging until I understand what's wrong.

Also note that I'm not using the authKey as mentioned in issue #159

### DIFF
--- a/enabler/src/de/schildbach/pte/NriProvider.java
+++ b/enabler/src/de/schildbach/pte/NriProvider.java
@@ -36,7 +36,7 @@ import okhttp3.HttpUrl;
  * @author Andreas Schildbach
  */
 public class NriProvider extends AbstractHafasLegacyProvider {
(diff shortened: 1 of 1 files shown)


schildbach:
Thanks, I merged this.

The new API endpoints seems to have different station IDs, so we need to adapt the unit test to reflect that. The problem with queryTrips seems to be a low-level problem on their side. Maybe we need to declare the provider can't do trips for now.

snake66:
Yes, I tried with the new Id's too. Same result. It's probably ok with no trip support for now, and I'll look into integrating their new api instead.

PR #161, opened 2017-05-26
Add Australia Navitia provider

### LINKED ISSUE #157: Reimplement PTV (Melbourne, Victoria) with new API
As identified in #124, the PTV department pulled their EFA feed, and is now encouraging people to use their new API. While we all argree this sucks, I'm going to go ahead and try to reimplement this API in public-transport-enabler so that we can get Melbourne back on the map.

I have a question regarding this. If someone has already written a [GPLv3 Java wrapper for this API](https://github.com/maccoda/PTV-API), can work with them to get it on Maven Central and use it as a dependency?

Or do you really prefer clean-room implementations for this project?

I realise that their library will not be a direct drop in replacement for `NetworkProvider`. However I would personally much rather implement a new `PtvNetworkProvider` with first class Java objects from their library, than to do more HTTP + JSON parsing.

pserwylo:
Navitia provides an "au" coverage which aggregates GTFS feeds from most states public transport authorities. At time of writing, it supports:

 * Melbourne
 * Sydney
 * Canberra
 * Brisbane
 * Darwin
 * Alice Springs
 * Perth
 * Adelaide
 * Hobart
 * Launceston
 * Bernie

However of these, Darwin, Alice Springs, and Darwin are out of date.

Also includes test coverage for each of the above datasets (except the out of date ones). By default only Melbourne is enabled though, to avoid running into API limits on Navitia if the tests are run too many times (e.g. via CI or often while developing).

Not quite sure if `setTimeZone("Australia/Melbourne");` is the right thing to do when Australia spans 3 different timezones.

Finally, had to make a minor change to the `AbstractNavitiaProvider` to allow the default `lineStyle()` method to get called for the Australian provider. This is because I needed to override so many of the styles for Melbourne. Most other providers have nice colours provided by Navitia. The `AbstractNavitiaProvider` does provide a method for overriding styles (`getLineStyle()`), but it doesn't pass through the `network` of the line. This is problematic because lines are often just named based on a number, which will quickly clash with a provider covering most major Australian cities.

Fixes #157 (sort of). Only sort of because it doesn't support realtime info from most of the Australian providers.

### DIFF
--- a/enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
+++ b/enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
@@ -140,27 +140,16 @@ public abstract class AbstractNavitiaProvider extends AbstractNetworkProvider {
 
     protected abstract String region();
 
-    private boolean preferNonNavitiaLineStyles = false;
-
-    /**
-     * Navitia is capable of returning "display_informations" which includes style related information. Depending on the
-     * coverage though, this may not be complete. Prefering non-Navitia line styles means we first delegate to
-     * {@link #lineStyle(String, Product, String)} before then falling back to {@link #getLineStyle(Product, String, String)}.
-     */
-    protected void setPreferNonNavitiaLineStyles(boolean preferNonNavitiaLineStyles) {
-        this.preferNonNavitiaLineStyles = preferNonNavitiaLineStyles;
-    }
-
     protected int computeForegroundColor(final String lineColor) {
         int bgColor = Style.parseColor(lineColor);
         return Style.deriveForegroundColor(bgColor);
     }
 
-    protected Style getLineStyle(final Product product, final String code, final String color) {
-        return getLineStyle(product, code, color, null);
+    protected Style getLineStyle(final String network, final Product product, final String code, final String color) {
+        return getLineStyle(network, product, code, color, null);
     }
 
-    protected Style getLineStyle(final Product product, final String code, final String backgroundColor,
+    protected Style getLineStyle(final String network, final Product product, final String code, final String backgroundColor,
             final String foregroundColor) {
         if (backgroundColor != null) {
             if (foregroundColor == null)
@@ -428,26 +417,12 @@ public abstract class AbstractNavitiaProvider extends AbstractNetworkProvider {
             final String code = displayInfo.getString("code");
             final String color = Strings.emptyToNull(displayInfo.getString("color"));
             final String name = Strings.emptyToNull(displayInfo.optString("headsign"));
-
-            final Style customLineStyle = getNonNavitiaLineStyle(network, product, code);
-            final Style lineStyle = customLineStyle != null ? customLineStyle : getLineStyle(product, code, color != null ? "#" + color : null);
+            final Style lineStyle = getLineStyle(network, product, code, color != null ? "#" + color : null);
 
             return new Line(lineId, network, product, code, name, lineStyle);
         } catch (final JSONException jsonExc) {
             throw new ParserException(jsonExc);
         }
-    }
-
-    /**
-     * @see #setPreferNonNavitiaLineStyles(boolean)
-     */
-    private Style getNonNavitiaLineStyle(String network, Product product, String code) {
-        if (preferNonNavitiaLineStyles) {
-            final Style style = lineStyle(network, product, code);
-            return style == Standard.STYLES.get(product) ? null : style;
-        }
-
-        return null;
     }
 
     private Stop parseStop(final JSONObject stopDateTime) throws IOException {
@@ -611,7 +586,7 @@ public abstract class AbstractNavitiaProvider extends AbstractNetworkProvider {
             final String name = Strings.emptyToNull(jsonLine.optString("name"));
             final String color = Strings.emptyToNull(jsonLine.getString("color"));
             final String textColor = Strings.emptyToNull(jsonLine.optString("text_color"));
-            final Style lineStyle = getLineStyle(product, code, color != null ? "#" + color : null,
+            final Style lineStyle = getLineStyle(network, product, code, color != null ? "#" + color : null,
                     textColor != null ? "#" + textColor : null);
 
             return new Line(lineId, network, product, code, name, lineStyle);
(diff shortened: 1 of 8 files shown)


pserwylo:
For the benefit of @losuler on #157, here is a devel version of Transportr running this Australian provider for testing:

[Transportr.Australian-dev-version.apk.zip](https://github.com/schildbach/public-transport-enabler/files/1031424/Transportr.Australian-dev-version.apk.zip)

schildbach:
This looks like a great PR, thanks!

@aelkhour @grote Do you have any comments on the change to `AbstractNavitiaProvider`?

We could add the network to `getLineStyle()` (EFA has something similar), but I'd prefer to change that first and then rebase this PR on the change. 

Clarification: All the live tests are not meant to be run by CI, some of them will fail anyway.

enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
```diff
@@ -417,14 +428,28 @@ else if (linkType.equals("physical_mode"))
...
             return new Line(lineId, network, product, code, name, lineStyle);
         } catch (final JSONException jsonExc) {
             throw new ParserException(jsonExc);
         }
     }
 
+    /**
+     * @see #setPreferNonNavitiaLineStyles(boolean)
+     */
+    private Style getNonNavitiaLineStyle(String network, Product product, String code) {
```
grote: Would be nice to annotate this `@Nullable`.

enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
```diff
@@ -140,6 +140,17 @@ public AbstractNavitiaProvider(final NetworkId network, final String authorizati
 
     protected abstract String region();
 
+    private boolean preferNonNavitiaLineStyles = false;
+
+    /**
+     * Navitia is capable of returning "display_informations" which includes style related information. Depending on the
+     * coverage though, this may not be complete. Prefering non-Navitia line styles means we first delegate to
+     * {@link #lineStyle(String, Product, String)} before then falling back to {@link #getLineStyle(Product, String, String)}.
+     */
```
grote: Couldn't we try to automatically take what is provided in `display_informations` and fall back to `lineStyle()` automatically when it isn't available? So we wouldn't need this method and class variable?

pserwylo: One problem is that sometimes `display_informations` is present, but not sufficient. For example, the state of Tasmania shows a green background for each bus, but the actual bus routes are coloured (I didn't override them in this PR, but will in the future).

However, I think perhaps @schildbach's suggestion of passing the network through to `getLineStyle()` would be nice. The thing is there is already a very sufficient implementation of `lineStyle()` in `AbstractNetworkProvider`, the one which works in conjunction with `setStyles()`. Over this weekend, I'll remove the hack I put in the Navitia provider, and instead pass through `network`. Then, there will be enough information to delegate to `lineStyles()` and all should be well. Thanks for the feedback.

enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
pserwylo: Agreed, but I think I'll be removing this, as per my comment above.

pserwylo:
Thanks for quick feedback. Unfortunately my response will not be as quick, but I'll try to address it over the coming days when I get a moment.

pserwylo:
Okay, 2b500e3 removes my dodgey hack and instead passes `network` through to `getLineStyles()`. This is much cleaner, so thanks for the suggestion. This is now ready for review again.

Qwertie-:
Any progress on this?

losuler:
I've been meaning to comment on this. I've been using the devel version @pserwylo posted a few comments above for a couple months now. 

I've been comparing it back and forth with the [PTV website journey planner](https://www.ptv.vic.gov.au/journey) and noticed most of my trips were mirroring what I saw on the site. The only discrepancy I noticed was that it appears Transportr is excluding a planned journey that the website isn't (but which still shows up in Transportr's departures). I'm not comfortable detailing my commute here, but I'm happy to talk in private if this wanted to be looked into.

Besides this, the only other thing I noted was how frequently and consistently trains are delayed (I'm not sure if this is the same for other lines). This just meant I was often looking at the website's real time departures, hence it would be nice to have this functionality in Transportr. However I understand this is probably out of the scope of this pull request.

schildbach:
Thanks @pserwylo, I squashed this into two commits and merged! Thanks also to @losuler, this kind of feedback is much appreciated.

I was thinking about not merging the RUN_EXPENSIVE_TESTS flag because actually all the *LiveTest tests are not included with CI (for the even more obvious reason that never would they all succeed at the same time). The `exclude 'de/schildbach/pte/live/**'` line in `build.gradle` is responsible for this. I ended up merged as is, but maybe we can agree on one of the concepts for excluding live tests from CI builds?

grote:
> The only discrepancy I noticed was that it appears Transportr is excluding a planned journey that the website isn't (but which still shows up in Transportr's departures). 

This could be a bug in Transportr. Feel free to open an issue there. Maybe you can reproduce this with a connection that does not disclose personal information.

pserwylo:
Thanks all. Apologies for not commenting, I didn't actually notice that it had been merged. I'm still using this version on my local device, and it is serving my purposes as a Melbournian who uses a lot of public transport (though usually only on the same regular routes).

> I ended up merged as is, but maybe we can agree on one of the concepts for excluding live tests from CI builds?

I have no objections to excluding from CI, as long as they can stay in the code. It did make me much more confident while developing to have the option to run the tests on an as-needed basis.

@schildbach feel free to ping me if anybody logs any issues with this provider, and I'll be happy to try and fix.

> Besides this, the only other thing I noted was how frequently and consistently trains are delayed (I'm not sure if this is the same for other lines). This just meant I was often looking at the website's real time departures, hence it would be nice to have this functionality in Transportr. However I understand this might be out of the scope of this pull request.

@losuler - thanks for the feedback. Indeed the first time I decided to contribute to this provider it was because I wanted to add support for realtime bus departures. However it was then I realised that PTV had pulled support for their old API used here. At that point, my job went from supporting realtime buses to "just making PTV work at all again" :) In the future when I get some more free time, I want to revisit the realtime buses (and trains) again. I think I'll create an issue and jot down some of my ideas there.

PR #167, opened 2017-09-05
Navitia: Add Orlyval to Physical Modes

grote:
Without this patch, connections with the orlyval train cause an IllegalArgumentException


schildbach:
Merged, thanks!

PR #169, opened 2017-10-01
[FR] NPE due to no color (train routes)

thony8:
In the GTFS used by Navitia, SNCF (French national train company) does not indicate colors for these routes, causing a Null Pointer Exception with the Style class.
The color is now defined when the network is "SNCF" in order to avoid this Exception (for FR-SW, SE and NE)

### DIFF
--- a/enabler/src/de/schildbach/pte/FranceNorthEastProvider.java
+++ b/enabler/src/de/schildbach/pte/FranceNorthEastProvider.java
@@ -49,34 +49,6 @@ public class FranceNorthEastProvider extends AbstractNavitiaProvider {
     }
 
     @Override
-    protected Style getLineStyle(final String network, final Product product, final String code, final String color) {
-        switch (product) {
-        case REGIONAL_TRAIN: {
-            // Rail (route_type = 2) for TER and Corail Intercité/Lunéa (all SNCF)
-            return new Style(Style.parseColor(color), computeForegroundColor(color));
-        }
-        case SUBURBAN_TRAIN: {
-            // Rail (route_type = 2) for Transilien (SNCF)
-            return new Style(Style.parseColor(color), computeForegroundColor(color));
-        }
-        case TRAM: {
-            // Tram (route_type = 0) for Strasboug (CTS) and Nancy (Stan)
-            return new Style(Shape.RECT, Style.parseColor(color), computeForegroundColor(color));
-        }
-        case BUS: {
-            // Bus (route_type = 3)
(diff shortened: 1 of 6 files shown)


enabler/src/de/schildbach/pte/FranceNorthEastProvider.java
```diff
@@ -53,11 +53,21 @@ protected Style getLineStyle(final String network, final Product product, final
...
+            return s;
         }
         case SUBURBAN_TRAIN: {
             // Rail (route_type = 2) for Transilien (SNCF)
-            return new Style(Style.parseColor(color), computeForegroundColor(color));
+            Style s;
+            if(network.equals("SNCF")) // SNCF data have no color
+                s =  new Style(Style.parseColor("#D90283"), computeForegroundColor("#D90283"));
+            else
+                s =  new Style(Style.parseColor(color), computeForegroundColor(color));
```
grote: What happens if some other network that is not SNCF also misses colors? Wouldn't it be better to have a more general situation and prevent crashes whenever a color is missing?

schildbach: Where does this #d90283 color stem from? Generally, if we don't know an "official" color/style for a line, we return null so that the app can decide on how to handle this (e.g. apply default color/style).

thony8: It is only an arbitrary color inspired by SNCF logo. There is a similar pull request to this problem.
It is just a patch to bypass this NPE and we can imagine the provider manage the default color and not only for the SNCF but I agree it is not perfect, this discussion is linked with #127

thony8:
I deleted the `getLineStyle()` override for FR-NE, FR-SW and FR-SE. It is better to let the application manages the default color. For example, it is the case for FR-NW between two towns with SNCF trains (SNCF doesn't provide colors in GTFS). IMO the NPE due to null colors is fixed.

grote:
@thony8 thanks for picking this up again! Something is odd though. This MR shows lots of unrelated commits and changes while it should only show your change. Maybe try rebasing it on latest master HEAD?

thony8:
Oops, sorry ... I fixed it.

arnaud-jacquemin:
Is there any way I can help on this issue? I'm willing to spend some time to help, because it would save me a lot of time if the SNCF was available in Transportr!

schildbach:
Finally merged this. Sorry for the delay!

PR #170, opened 2017-10-16
Add GhanaProvider

grote:
This is similar to #146 in the sense that this work came out of a local initiative to map public transport data in a region where otherwise no information and no apps exist. In Accra, the public transport is even informal without fixed schedules. The data was entered into OpenStreetMap and a GTFS file generated with [osm2gtfs](https://github.com/grote/osm2gtfs). This file is then processed by Navitia that offers the routing server for the data.

mapping party

This PR is mostly for inclusion in Transportr. It is not a request for adding Ghana support to Oeffi.

### DIFF
--- a/enabler/src/de/schildbach/pte/GhanaProvider.java
+++ b/enabler/src/de/schildbach/pte/GhanaProvider.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2014-2015 the original author or authors.
+ *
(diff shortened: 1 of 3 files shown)


nlehuby:
Hi,

any update on that ? 
We are looking forward to test TransportR in Accra.

schildbach:
Merged this PR.

nlehuby:
thanks !

PR #173, opened 2017-11-08
Navitia: Fix Sorting Order of Suggested Locations and Address Names

grote:
The navitia developers have informed me that the "quality" of
auto-completion results is deprecated and already meaningless.
Also, we had a wrong inverse sorting order before. The only thing that
matters is the sort order in which the results are returned.
This commit fixes the sort order accordingly.

Similiarly broken was the way we displayed addresses. The house number
was missing from the location name. This commit adds the house number
behind the street name if available and gives a way to define other
formats like in France where the house number comes before the street
name.

Last but not least, some `@Nullable` annotations were added to modified
classes to highlight existing null pointer problems and to make it
easier for future developers to become aware of them.

### DIFF
--- a/enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
+++ b/enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
@@ -208,6 +208,19 @@ public abstract class AbstractNavitiaProvider extends AbstractNetworkProvider {
         return name;
     }
 
+    /**
+     * Navitia always formats the address label in the French way.
+     * This method can be used to display addresses in a localized way.
+     * It defaults to putting the house number behind the street name.
+     *
+     * @param name The name of the address. Usually a street name.
+     * @param houseNumber The house number of the address.
+     * @return the localized name of the address location
+     */
+    protected String getAddressName(final String name, final String houseNumber) {
+        return name + " " + houseNumber;
+    }
+
     private Location parsePlace(JSONObject location, PlaceType placeType) throws IOException {
         try {
             final LocationType type = getLocationType(placeType);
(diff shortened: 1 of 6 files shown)


enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
```diff
@@ -17,7 +17,12 @@
 
 package de.schildbach.pte;
 
-import static com.google.common.base.Preconditions.checkNotNull;
+import com.google.common.base.Strings;
```
schildbach: Re-organizing imports causes noisy commits. Is this necessary?

enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
```diff
@@ -208,6 +207,20 @@ protected String getLocationName(String name) {
         return name;
     }
 
+    /**
+     * Navitia always formats the address label in the French way.
+     * This method can be used to display addresses in a localized way.
+     * It defaults to putting the house number behind the street name.
+     *
+     * @param name The name of the address. Usually a street name.
+     * @param houseNumber The house number of the address. "0" if none available.
```
schildbach: Personally I'd prefer to handle the '0' special case outside of this method and use `null` instead.

enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
grote: Android Studio does this automatically for me :(
I'll remove them manually.

enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
grote: How about just calling this method if there is a non `null` or `"0"` housenumber?

schildbach: Would be fine to me too.

schildbach: All I want to avoid is bleed a broken API (use '0' if you mean 'none') into our own code.

grote:
I have changed the commit of this PR to be only about the missing house numbers. Another one for the sort order will follow shortly.

schildbach:
Merged, thanks!

PR #174, opened 2017-11-08
Navitia: Fix sort order of suggested locations

grote:
The navitia developers have informed me that the "quality" of
auto-completion results is deprecated and already meaningless.
Also, we had a wrong inverse sorting order before. The only thing that
matters is the sort order in which the results are returned.
This commit fixes the sort order accordingly.

### DIFF
--- a/enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
+++ b/enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
@@ -845,10 +845,11 @@ public abstract class AbstractNavitiaProvider extends AbstractNetworkProvider {
 
             if (head.has("places")) {
                 final JSONArray places = head.getJSONArray("places");
+                final int numPlaces = places.length();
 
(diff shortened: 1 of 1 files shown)


enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
```diff
@@ -845,10 +845,11 @@ public SuggestLocationsResult suggestLocations(final CharSequence constraint) th
 
             if (head.has("places")) {
                 final JSONArray places = head.getJSONArray("places");
+                final int priorityBase = places.length();
 
                 for (int i = 0; i < places.length(); ++i) {
```
schildbach: Just a small thing, but you could use the introduced variable here too. And maybe `numPlaces` would be a more general name?

grote: Fixed

schildbach:
Thanks for spotting the problem with inverse sort order. Merged!

PR #175, opened 2017-11-12
added Basel Verkehrsbetriebe (BVB)

culmat:


schildbach:
I had removed that with commit 90b0cc02cd3460fc388b1304c900324bc5e51cce because it seemed very out of date and Basel is covered by other providers too. Why are you adding it back?

culmat:
Hi Andreas,
I was not aware it had been there before. Which provider should I use for BVB? 
I don't think it is out of date. It is the back end of the [official BVB App](https://play.google.com/store/apps/details?id=com.mdv.BVBCompanion) with > 100.000 installations.

Regards
Matthias

schildbach:
Ah ok I see they're using EFA-BW. However that URL is no more since a few days – does BVBs app still work?

schildbach:
I'm closing this, as the URL doesn't exist. If you've got an update, feel free to comment and I'll re-open.

PR #178, opened 2017-12-05
Navitia: Add New Zealand Provider

grote:
Currently supports Wellington, Auckland and Christchurch.

This provider has been in use with Transportr already. Time to upstream it.


schildbach:
Merged, thanks!

PR #180, opened 2017-12-05
Navitia: Add Spain Provider

grote:
This was used in Transportr for some time and I used in it Valencia a couple of times.

The usual issues with Navitia based providers that cover a large region apply:
* quality of auto-complete results is bad
* sometimes validity of sub-regions expire and stop working


schildbach:
Merged, thanks!

PR #185, opened 2018-01-04
KVV: new endpoint

alexander-albers:
It seems like the KVV provider doesn't allow any requests through its old endpoint anymore, as all requests result in `400: Bad Request`. Instead, I found that the official website now uses the following url: [https://www.kvv.de/tunnelEfaDirect.php](https://www.kvv.de/tunnelEfaDirect.php) with the `action` query parameter set to the respective request method (eg. `XSLT_TRIP_REQUEST2` for trip requests). The host and first query parameter were the only things that had to be changed, all other requests parameters can be left untouched. Also, the responses seem to be the same.
I suspect that the original (ip-address) endpoint now only allows requests from the official server so that all _user_ requests can be encrypted using https.

### DIFF
--- a/enabler/src/de/schildbach/pte/KvvProvider.java
+++ b/enabler/src/de/schildbach/pte/KvvProvider.java
@@ -34,14 +34,17 @@ import okhttp3.HttpUrl;
  * @author Andreas Schildbach
  */
 public class KvvProvider extends AbstractEfaProvider {
-    private static final HttpUrl API_BASE = HttpUrl.parse("http://213.144.24.66/kvv2/");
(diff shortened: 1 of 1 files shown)


alexander-albers:
Small update: the url path has likely been migrated to [http://213.144.24.66/sl3/](http://213.144.24.66/sl3/). Nevertheless, I still think it would be better to use the new https endpoint as provided above.

schildbach:
Sorry for not updating this PR in time. The SSL certificate of your IP pointed to projekte.kvv-efa.de to I chose to use that one instead.

But thanks anyway for your help in this matter, it's much appreciated!

PR #99, opened 2016-06-10
Add Texas, USA provider

oja:
Provider for Houston, Dallas, San Antonio, Fort Worth, and Austin.

I tried to run gradle tests, but came up with some errors I suspect were caused due to not running with proper android support. Please let me know if I messed anything up.

PR #194, opened 2018-05-12
Add Texas, USA provider

oja:
This PR is a resubmit of #99 with fixed conflicts.

Navitia `us-tx` includes support for Austin (CapMetro), Dallas (Dallas Rapid Transit), Fort Worth (FWTA), Houston (Harris County Metro), and San Antonio (VIA).

### DIFF
--- a/enabler/src/de/schildbach/pte/NetworkId.java
+++ b/enabler/src/de/schildbach/pte/NetworkId.java
@@ -76,7 +76,7 @@ public enum NetworkId {
     DUB,
 
     // United States
-    RTACHICAGO, OREGON,
+    RTACHICAGO, OREGON, TEXAS,
 
     // Canada
     ONTARIO, QUEBEC, BRITISHCOLUMBIA,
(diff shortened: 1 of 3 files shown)


grote:
Did you use this yourself already and can tell whether it works well there?

oja:
@grote From what I can tell, Google Maps uses GTFS static + realtime for Austin, and I've never had a problem. Navitia sources CapMetro from [here](https://data.texas.gov/Transportation/GTFS/r4v4-vz24) and routes seem to be updated regularly (last was early April). I think Navitia does not contain realtime, however.

I can't confirm Houston, DFW, or San Antonio. Although I would be very surprised if it were worse than Austin, given that they're larger cities.

schildbach:
We now have CmtaProvider which should cover Austin + Texas. It's Hafas-based, so I have a bit more confidence into it than using Navitia. What do you think?

Tavlor:
I can confirm that the CmtaProvider only covers Austin. I'm interested in adding support for Dallas in particular; is it better to split Texas up by city or keep it as it is here?

schildbach:

PR #199, opened 2018-06-23
Added Canberra, Australia Provider

Chickensoupwithrice:
This PR adds support for the bus network in Canberra Australia. 
Navitia includes support for the Transport Canberra network, which covers all of Canberra's buses (currently the only available public transport network)

I have not tested this yet, because I'm not entirely sure how to do that. This pull request was based on the Texas USA provider #194

### DIFF
--- a/enabler/src/de/schildbach/pte/AustraliaProvider.java
+++ b/enabler/src/de/schildbach/pte/AustraliaProvider.java
@@ -32,6 +32,7 @@ public class AustraliaProvider extends AbstractNavitiaProvider {
     public static final String NETWORK_QLD = "TransLink";
     public static final String NETWORK_SA = "Adelaide Metro";
     public static final String NETWORK_WA = "Transperth";
+    public static final String NETWORK_ACT = "ACTION";
 
     private static final Map<String, Style> STYLES = new HashMap<>();
 
--- a/enabler/src/de/schildbach/pte/CanberraProvider.java
+++ b/enabler/src/de/schildbach/pte/CanberraProvider.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright the original author or authors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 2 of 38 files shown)


enabler/src/de/schildbach/pte/AustraliaProvider.java
```diff
@@ -32,6 +32,7 @@
     public static final String NETWORK_QLD = "TransLink";
     public static final String NETWORK_SA = "Adelaide Metro";
     public static final String NETWORK_WA = "Transperth";
+    public static final String NETWORK_ACT = "ACTION";
```
grote: What's that for?

enabler/src/de/schildbach/pte/CanberraProvider.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * @author Anish Lakhwara
+ */
+
+public class CanberraProvider extends AbstractNavitiaProvider {
+    private static String API_REGION = "au";
+
+    public TexasProvider(final HttpUrl apiBase, final String authorization) {
+        super(NetworkId.TEXAS, apiBase, authorization);
```
grote: Using `TEXAS` here won't work. You need to use your own `NetworkId`.

enabler/src/de/schildbach/pte/CanberraProvider.java
```diff
@@ -0,0 +1,43 @@
...
+        setTimeZone("Australia/Canberra");
+    }
+
+    public CanberraProvider(final String authorization) {
+        super(NetworkId.TEXAS, authorization);
```
grote: Same here.

enabler/src/de/schildbach/pte/AustraliaProvider.java
Chickensoupwithrice: The Australia Provider didn't originally include the Canberra network, which is already a part of the ```au``` dataset on Navitia

grote: But this constant isn't used by anything. Also, you are adding a new provider it seems rather than modifying an existing one.

grote: In this case, it might very well be. Did you try to use the Australia provider in Transportr and check if it returns canberra results?

Chickensoupwithrice: Ahhh, yep it's working. I just tried the 2.0.0rc3 and it seems to be working just fine from preliminary searches. Is there any way for me to check all routes in Canberra? 

I think this PR is useless and can be closed now. Thanks for your help! :)

grote: > Is there any way for me to check all routes in Canberra?

No, PTE (this library) has no API for that.

> I think this PR is useless and can be closed now. Thanks for your help! :)

You should be able to close it yourself.

PR #205, opened 2018-07-21
Define defaultStyle for unknown products

mimi89999:


schildbach:
Thanks to #206 this case should now not happen any more, unless we add products and forget about defaults. I think I'd prefer an exception at this point.

mimi89999:
OK

PR #206, opened 2018-07-21
Add default background color for ON_DEMAND

### LINKED ISSUE #204: Missing default background color for ON_DEMAND
There is no default color for `ON_DEMAND` that is causing an NPE when Navitia doesn't return a color for that product. I can add it, but I don't know what color to choose.

@grote @schildbach

mimi89999:
Closes #204


schildbach:
Merged.

PR #216, opened 2018-08-07
Implement TripOptions for queryTrips method

### LINKED ISSUE #214: Move most of the parameters for queryTrips in new TripOptions class
This is just a follow up on [#102 (comment)](https://github.com/schildbach/public-transport-enabler/pull/102#issuecomment-407667057) to discuss the idea.

I've had a quick look which of the parameters could be moved into a new `TripOptions` class. I would leave `from`, `to` and `date` as normal parameters, while moving all the others into the new class (`dep`, `products`, `optimize`, `walkSpeed`, `accessibility`, `options`) because they seem quite "auxiliary". Would you be ok with this distribution @schildbach ?

ialokim:
This PR moves the following "auxiliary" parameters from `queryTrips` into a separate `TripOptions` class while leaving `from`, `via`, `to`, `date` and `dep` as normal parameters: `products`, `optimize`, `walkSpeed`, `accessibility` and `options`.
See also the quick discussion in #214.

For this, I've added a new method signature in the `NetworkProvider` interface while marking the old one as _deprecated_. I've also changed the implementations in all the subclasses to use the new parameters and implemented the old method signature in `AbstractNetworkProvider` as a simple alias for the new one.

I have run all the live tests to make sure the PR does not break any providers. There were exactly the same amount of failures before and after the changes, but I was not able to test providers requiring secret keys except the Navitia based ones.

Fixes #214.

### DIFF
--- a/enabler/src/de/schildbach/pte/NetworkProvider.java
+++ b/enabler/src/de/schildbach/pte/NetworkProvider.java
@@ -68,6 +68,43 @@ public interface NetworkProvider {
         BIKE
     }
 
+    public class TripOptions {
+        public final @Nullable Set<Product> products;
+        public final @Nullable Optimize optimize;
+        public final @Nullable WalkSpeed walkSpeed;
+        public final @Nullable Accessibility accessibility;
+        public final @Nullable Set<Option> options;
+
+        /**
+         * @param products
+         *            products to take into account, or {@code null} for the provider default
+         * @param optimize
+         *            optimize trip for one aspect, e.g. duration
+         * @param walkSpeed
+         *            walking ability, or {@code null} for the provider default
+         * @param accessibility
+         *            route accessibility, or {@code null} for the provider default
+         * @param options
+         *            additional options, or {@code null} for the provider default
+         */
+        public TripOptions(@Nullable Set<Product> products, @Nullable Optimize optimize, @Nullable WalkSpeed walkSpeed,
+                           @Nullable Accessibility accessibility, @Nullable Set<Option> options) {
+            this.products = products;
+            this.optimize = optimize;
+            this.walkSpeed = walkSpeed;
+            this.accessibility = accessibility;
+            this.options = options;
+        }
+
+        public TripOptions() {
+            this.products = null;
+            this.optimize = null;
+            this.walkSpeed = null;
+            this.accessibility = null;
+            this.options = null;
+        }
+    }
+
     NetworkId id();
 
     boolean hasCapabilities(final Capability... capabilities);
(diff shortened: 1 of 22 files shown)


enabler/src/de/schildbach/pte/NetworkProvider.java
```diff
@@ -68,6 +68,38 @@
         BIKE
     }
 
+    public class TripOptions {
+        public @Nullable Set<Product> products;
```
schildbach: Can we make these fields final?

enabler/src/de/schildbach/pte/NetworkProvider.java
```diff
@@ -160,8 +188,8 @@ QueryTripsResult queryTrips(Location from, @Nullable Location via, Location to,
      * 
      * @param context
      *            context to query more trips from
-     * @param next
```
schildbach: This fix seems unrelated to the PR. Please don't mix.

enabler/src/de/schildbach/pte/VrsProvider.java
```diff
@@ -688,6 +686,11 @@ public QueryTripsResult queryTrips(final Location from, final @Nullable Location
                     QueryTripsResult.Status.UNKNOWN_TO);
         }
 
+        if (options == null)
+            options = new TripOptions();
+        if (options.products == null)
```
schildbach: These two lines also seem to be unrelated, but let me know if I'm wrong.

grote:
> I wonder if we perhaps should rename the current Options enum to TripFlags to make the API clearer.

Good idea!

schildbach:
Then again if we plan the Options to TripFlags rename in this case the API will break anyway so there would be not much point backwards compatibility in this PR.

enabler/src/de/schildbach/pte/NetworkProvider.java
ialokim: Ok, I'll change that.

enabler/src/de/schildbach/pte/NetworkProvider.java
ialokim: Sure, but then we would have to change some logic where e.g. `options.products` is excepted to be non-null (see [here](https://github.com/schildbach/public-transport-enabler/pull/216/files/13e16d3e4571d96572828b2388dfaeb058aefe6e#diff-b2543011aff601f59664df7f7e17430cR851) and some lines downwards as an example).

Do you want me to amend to this commit or add a new commit with these changes or do you want to change it while merging?

enabler/src/de/schildbach/pte/VrsProvider.java
ialokim: Hum they are kind of related, because in line 700 (old file) it expected the `products` to be non-null, so that's what I wanted to ensure here.

schildbach: I understand the problem, and I agree the NPE should be fixed. It's just not related to this PR, as the bug has existed before.

I just fixed the bug in master via 458fea99ac0bbe4dc8a526bb9e98d8c4f3eabb88. If you rebase your PR on master the need for this specific change should go away.

enabler/src/de/schildbach/pte/NetworkProvider.java
schildbach: Yes, please do. You can fix it in a separate PR if you want, then I'll merge the JavaDoc fix right away.

enabler/src/de/schildbach/pte/NetworkProvider.java
schildbach: Oh yes, indeed. Please do not mutate the TripOptions instance. It's owned by the caller, who will not expect it to be changed as a side effect of calling.

ialokim:
Thanks @schildbach for your comments.

I've now rebased the PR on current master, changed the `TripOptions` fields to final (I've also had to change the `products` handling in the `NegentweeProvider.java`) and split out the javadoc fix into #217.

schildbach:
I merged this, many thanks! I decided to move TripOptions to the .dto package.

PR #217, opened 2018-08-10
Fix javadoc for queryMoreTrips

ialokim:
Just a little fix in the javadoc of this method. See #216


schildbach:
Good catch, thanks! I found another instance of this in HslProvider. I added that to the commit and merged.

Issue #224, opened 2018-09-14
BC navita provider: no PhysicalMode "NA"


anjandev:
![screenshot_transportr_20180912-000845](https://user-images.githubusercontent.com/8602360/45532764-15a10e80-b7aa-11e8-91e3-d85f3a2ea805.png)

anjandev:
![screenshot_transportr_20180912-000845](https://user-images.githubusercontent.com/8602360/45532764-15a10e80-b7aa-11e8-91e3-d85f3a2ea805.png)

anjandev:
I think  I fixed it here:

https://github.com/anjandev/public-transport-enabler/commit/b8ec92de14515d452b8230fe75458c0958d9065e

anjandev:
I made a mistake. I will open a new issue with everything proper. My bad

anjandev:
I investigated this issue. Turns out that when the route includes the skytrain (the overhead subway in vancouver) Navitia returns the "NA" physical mode. 

https://api.navitia.io/v1/coverage/ca-bc/journeys?from=poi%3Aosm%3Anode%3A384252960&to=poi%3Aosm%3Away%3A102254132&

schildbach:
What does "NA" stand for? Maybe "not available" or "not applicable"? If so, then I'd vote for removing the enum value entirely and use "null" instead.

anjandev:
@schildbach in this case, vancouver's skytrain was placed as NA. Should I put it as null? Would that not just remove the physical mode from apps? Ie. When a route requires the skytrain it's just a blank?

schildbach:
Actually I'm not so sure any more. PhysicalMode is used within AbstractNavitiaProvider only, so apps (= users of the library) won't care about it.

@aelkhour once introduced the enum in commit 9f2b3b57e126864ec8de5dc8c7cee29e8606a078, maybe he want to add his view to the discussion?

anjandev:
@schildbach I sent an email to @aelkhour and I have gotten no response. It has been a week. 

What is the purpose of the physical_mode. Like what is it used for? Can we change it to Null?

schildbach:
@prhod This issue about "NA" PhysicalMode in the British Columbia dataset is very similar to #236. What does NA stand for? Is it perhaps an error in the dataset?

schildbach:
I should mention that according to http://doc.navitia.io/#public-transport-objects NA is not valid.

grote:
[...]

PR #226, opened 2018-09-21
Navitia added NA physical mode for skytrain (Subway in BC)

anjandev:


anjandev:
fixes #224

anjandev:
Torsten Grote <notifications@github.com> writes:

> @grote commented on this pull request.
>
> ----------------------------------------------------------------------------------------------------
>
> In enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java:
>
>> @@ -621,6 +621,8 @@ private Line parseLine(final JSONObject jsonRoute) throws IOException {
>  
>          switch (physicalMode) {
>          case BUS:
> +        case NA:
> +            return Product.SUBWAY;
>
> Not sure it makes sense to add this for all Navitia providers. What does NA stand for? If it means "not available", it
> might not always be subway.
>
> Also, the Skytrain in Vancover is more SUBURBAN_TRAIN or TRAM than SUBWAY I'd argue.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub, or mute the thread.

Ya, I think NA stands for not applicable. I couldnt find anything on the
Navitia website about it. I can add it to only the BC provider rather
than all Navitia providers.

I thought the skytrain was also more SUBURBAN_TRAIN but I check the
wikipedia page on subway:

https://en.wikipedia.org/wiki/Rapid_transit

and it stated that the transport system might be called the subway even
though it's primarily over ground (search for "skytrain") on this page.

I can change to SUBURBAN_TRAIN as that makes more sense.

Let me know what I should change to get this fixed.

anjandev:
It seems that australia also has a `NA` physical mode:

http://canaltp.github.io/navitia-playground/play.html?request=https%3A%2F%2Fapi.navitia.io%2Fv1%2Fcoverage%2Fau%2Fphysical_modes%2Fphysical_mode%253ANA%2F%3F&token=6eb3adff-51bf-4e77-bb4b-508c09e91778


I will send a seperate PR that makes this code specific to the British Columbia Provider.

schildbach:
Is this still `NA` physical mode still an issue? @prhod wanted to work on getting rid of those in the original data.

prhod:
As stated in [this comment](https://github.com/schildbach/public-transport-enabler/issues/234#issuecomment-578662019), there should be no more issues like that.

schildbach:
Ok, so I'll close this.

PR #228, opened 2018-09-23
Negentwee: use the correct type parameter name

full-duplex:
Previous fix accidentally changed the query parameter name, causing it to be ignored.

This changes it back and tests the parameter.


schildbach:
Merged.

PR #233, opened 2018-10-07
Negentwee: Improve date and time parsing and add timezone awareness

### LINKED ISSUE #227: Negentwee departure dates are off by ~50 years
That's because 9292.nl departure boards only shows the time (not the date) and we parse it as 1970-01-01. The problem is even worse on departure boards that wrap over the date ... the API doesn't make it clear that the early morning times are in fact on the next day.

@full-duplex Do you know if that problem is already discussed at 9292? IMHO ideally they would *always* use ISO 8601 (incl. date and TW) and be done with it.

full-duplex:
- Extend and use the ParserUtils functions instead and assume parsed time strings are within the next 24 hours instead of leaving the date at 1970-01-01 (closes #227).
- Use the Europe/Amsterdam timezone when formatting and parsing server dates.

### DIFF
--- a/enabler/src/de/schildbach/pte/NegentweeProvider.java
+++ b/enabler/src/de/schildbach/pte/NegentweeProvider.java
@@ -19,10 +19,10 @@ package de.schildbach.pte;
 
 import java.io.IOException;
 import java.io.Serializable;
-import java.text.ParseException;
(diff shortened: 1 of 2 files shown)


schildbach:
Merged, thanks!

PR #254, opened 2019-01-04
VgsProvider: Add Salt and enable Mic-Mac

ialokim:
For the new saarfahrplan.de endpoint to work, it requires the mic-mac parameters to be send. This PR adds the salt found by @juliuste and @derhuerst from their [hafas-client](/public-transport/hafas-client/blob/master/p/saarfahrplan/index.js#L39).

### DIFF
--- a/enabler/src/de/schildbach/pte/VgsProvider.java
+++ b/enabler/src/de/schildbach/pte/VgsProvider.java
@@ -19,6 +19,8 @@ package de.schildbach.pte;
 
 import java.util.regex.Matcher;
 
+import com.google.common.base.Charsets;
+
 import de.schildbach.pte.dto.Product;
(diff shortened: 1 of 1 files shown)


schildbach:
Thanks, but is the salt currently really needed? I'd say not, because the unit tests work even without it.

I'll keep your PR open though, just in case.

ialokim:
> unit tests work even without it.

That is quite strange, they all fail for me using the current `master` branch with `java.lang.RuntimeException: AUTH`.

schildbach:
Interesting. Seems as if the mic/mac salt is dependent on the aid. I've merged this for now since it doesn't hurt, but maybe in future we need to make the salt properly configurable.

PR #258, opened 2019-02-09
Fix live tests for departures with equivs

ialokim:
The live tests fail for cases where equivalent stations can be queried for departures too, but only one station is actually returned. The live test should accept having `result.stationDepartures.size() == 1` for these cases.


schildbach:
Yes that's true. But the idea was that you pick stations that actually return multiple stations. Just as you would pick stations that do return any results at all.

ialokim:
Okay, got it. Thanks for your explanation! :+1:

PR #261, opened 2019-02-21
Fix AbstractNavitiaProvider to use UTC times

### LINKED ISSUE #260: AbstractNavitiaProvider: Use UTC times
As stated in grote/Transportr#577 by @schildbach:
> the API is UTC time for both request and response

This is currently not implemented for `AbstractNavitiaProvider` and therefore neither for all of the several providers using Navitia's API. The necessary changes would be in [line 320](https://github.com/schildbach/public-transport-enabler/blob/master/enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java#L320) and 324, respectively.

ialokim:
This PR fixes current problems when sending a request to Navitia from within another timezone as the provider has. Until now, the date and time is sent to Navitia as if local and provider's timezone were equal. By setting `DateFormat.timeZone` to the provider's one while converting time to string in both directions, it now works as expected.

So `AbstractNavitiaProvider` now is in line with the expected behavior stated by @schildbach in grote/Transportr#577:

> the API is UTC time for both request and response

Fixes #260.

### DIFF
--- a/enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
+++ b/enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java
@@ -20,6 +20,7 @@ package de.schildbach.pte;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import java.io.IOException;
+import java.text.DateFormat;
 import java.text.ParseException;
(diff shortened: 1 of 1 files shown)


schildbach:
Thanks for spotting this! Merged.

PR #271, opened 2019-07-08
Select provider for pte service from configuration file

xamanu:
This PR introduces a `config.json` file, which must be similar to the `example.config.json` file. It defines the  service's provider (and a token if required, or can be omitted).

Not sure whether this is the best/desired approach. Perhaps it would be better to handle the authorization tokens similar and together with the `secrets.properties`!? But shouldn't live this more globally then? And do we want to include the provider selector in this file, in order to get rid of the json file? @schildbach, some guidance would be really nice.


xamanu:
Work is continuing in https://gitlab.com/oeffi/public-transport-service Closing here.

PR #276, opened 2019-07-29
Add prague provider

filip-hejsek:


schildbach:
Thanks for your contribution! I'm sorry for the delay.

Does the "cz" region include only Prague? Theoretically, it should be the entire Czech Republic?

schildbach:
Apart from the above question, the PR looks very good to me.

filip-hejsek:
Well, right now it's only Prague because data for other cities aren't available in gtfs, though in the future other cities might be added.

schildbach:
Thanks. If you don't mind I'll will then rename your provider to CzechRepublicProvider (and the NetworkId constand accordingly). Because it's likely Navitia will add other Czech cities under the cz region as well.

filip-hejsek:
Yeah, it's probably better to rename it to CzechRepublicProvider.

schildbach:
Renamed & merged.

PR #283, opened 2019-10-15
Enable Java 8 language features.

schildbach:
Note that for Android compatibility, we still need to stay compatible to the Java 7 API.


schildbach:
Merged.

PR #287, opened 2019-11-16
Line colors for Hagen (VRR)

d3d9:
Added newly introduced line colors for Hagen.

~~For some reason, the previous HST (Hagener Straßenbahn AG) network is now named "x19" instead of "hst", so the hst colors will only be visible starting December 15th (because~~ the new network has the correct name "hst"~~)~~.

_Public source: <http://mehr-fuer-hagen.de>_


schildbach:
Merged! Thanks for the contribution.

PR #294, opened 2020-01-02
Negentwee: Use full trip fares instead of per leg

full-duplex:
The current implementation adds the individual fares for each leg to the Trip object. Looking at the other providers, this is not correct, and there is no way to express individual leg fares. Transportr also assumes the fares are for the full trip, so it displays the wrong price.

I removed the per leg fares for now and only add the full-price and the reduced-price fares for the whole trip if available (there is no way to express the difference between the two with the current fare types, other than the name string).


schildbach:
Merged, thanks!

PR #295, opened 2020-01-03
AbstractNavitiaProvider: add coverage to url for trip queries

### LINKED ISSUE #275: AbstractNavitiaProvider: use full url for TripQuery
As pointed out (some time ago) in [this issue](https://github.com/grote/Transportr/issues/583#issuecomment-466106840), `AbstractNavitiaProvider` does not add `coverage` information into the url when searching for a trip (on endpoint `journeys`), see [here](https://github.com/schildbach/public-transport-enabler/blob/master/enabler/src/de/schildbach/pte/AbstractNavitiaProvider.java#L902).

This does not cause any problems when there is no error as Navitia can handle queries without the `coverage` part. But in the case of an invalid station, it seems to make a difference: Navitia's returning a `BlockedException` instead of the `invalid_station` flag.

I guess this could be solved by simply adding the `coverage` part into the url as done in all the other methods (e.g. `suggestLocations`). They use the method `url()` which returns the base path including the `coverage`.

If you want me to do so, I could make the necessary change and open an PR.

ialokim:
As already discussed in #275, I've changed the AbstractNavitiaProvider to include the `coverage` part in the trip query.

Fixes #275.

### DIFF
--- a/src/de/schildbach/pte/AbstractNavitiaProvider.java
+++ b/src/de/schildbach/pte/AbstractNavitiaProvider.java
@@ -893,7 +893,7 @@ public abstract class AbstractNavitiaProvider extends AbstractNetworkProvider {
 
         try {
             if (from != null && from.isIdentified() && to != null && to.isIdentified()) {
-                final HttpUrl.Builder url = apiBase.newBuilder().addPathSegment("journeys");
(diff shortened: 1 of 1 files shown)


schildbach:
Merged.

Issue #183, opened 2017-12-21
efa.ivb.at (Innsbruck/IVB): API gone (404)


girst:
The IVB API has been removed; any requests to `https://efa.ivb.at/ivb/` are redirected to a 404 page (and the subdomain itself now returns the homepage). 

The new web interface for trip requests is at `https://fahrplan.ivb.at`, but I don't think a public API documentation is available; the usual paths (`/ivb/XSLT_TRIP_REQUEST2` etc.) don't exist there any more.

girst:
just poking around a bit: requests seem to be handled by `/bin/mgate.exe` -- a quick google search returns entries about DB's HAFAS system; maybe this API can be adapted (although I personally lack the know how about android programming)

schildbach:
Public-transport-enabler is a standard Java library – no Android involved. If this is Hafas we can likely talk to it.

schildbach:
I removed IvbProvider via ea0865f3c236a70ff20196378454e0ccd394a895 as their EFA API is gone forever. I think it's safe to assume VvtProvider can be used in place as that one covers Tyrol (which Innsbruck belongs to).

simon04:
`VvtProvider` is disabled in Öffi "because the VVT explicitly desires not to be included in Öffi". :disappointed:

schildbach:
@simon04 Yes, sadly.

PR #312, opened 2020-06-24
VVT: EFA-based provider

girst:
I've stumbled upon this endpoint today. It would be a suitable replacement for Innsbruck/IVB (#183). This API is provided by VVT themselves, not through VAO, who according to @schildbach are opposing the use of "their" data, so this may be a possibility to get VVT back into Öffi. (the domain points directly to a VVT-owned IP, not to VAO)

The API only returns results for Tirol, meaning suggestions will be better than those provided by OebbProvider or the VAO based VvtProvider.

I wasn't sure if it's appropriate to replace the current `VvtProvider` (the blacklisted VAO one, which returns the same results as VaoProvider), so I called it `VvtEfaProvider`. Also, I added my name to the VvtEfa* files, but I doubt my changes are even significant enough to be copyrightable, so I have no problem to remove the mention.

Also pinging @grote, since I'd love to see this in Transportr.

Commit message:
>This essentially reverts ea0865f3c236a70ff20196378454e0ccd394a895,
>updates the endpoint URL, renames the class and fixes the tests (the
>station ids changed from 604xxxxx to 601xxxxx).
>
>This endpoint is officially used by VVT to generate PDF timetables; the
>domain points to a VVT (not VAO) owned IP/server.


schildbach:
I wonder how much future has this provider got? What's the data quality, compared to their Hafas/VAO-backed API? Is the EFA API used by any of their own frontends/apps?

girst:
On Sat, Jun 27, 2020 at 12:35:10AM -0700, Andreas Schildbach wrote:
>I wonder how much future has this provider got?

I don't think that API will get removed very soon, but of course there's 
no knowning.

>What's the data quality, compared to their Hafas/VAO-backed API?

No realtime data, otherwise not bad. I can't currently compare with VAO, 
since Transportr raises an AUTH runtimeexception (opening ticket in a 
second), but it looks like it's all there.

>Is the EFA API used by any of their own frontends/apps?

yes, for creating pdfs of time tables. here is an example: 
https://vvt.at/page.cfm?vpath=fahrplan/linienfahrplan&stop_id=60101187&line=ivb:01002:P:R:j20&i=Innsbruck%2C%20Hauptbahnhof%20%E2%80%93%20Tram%202

PR #333, opened 2020-09-13
Add new VIA and REALTIME capabilities

### LINKED ISSUE #95: New Capability: VIA and REALTIME

ialokim:
This PR adds three new capabilities for network providers:
* `TRIPS_VIA` for providers that allow specifying a location to pass by
* `DEPARTURES_REALTIME` for providers that support real-time data on departures
* `TRIPS_REALTIME` for providers that support real-time data on trip queries.

As discussed in #95, the new capabilities are split into different commits.

Fixes #95.

---

This is a copied and rebased version of #210 after the source repository has been deleted. Sorry for the noise here.

@schildbach Would be nice to have this merged, if you have no further objections.

### DIFF
--- a/src/de/schildbach/pte/AbstractEfaProvider.java
+++ b/src/de/schildbach/pte/AbstractEfaProvider.java
@@ -23,6 +23,7 @@ import static com.google.common.base.Preconditions.checkState;
 import java.io.IOException;
 import java.io.Reader;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Calendar;
 import java.util.Currency;
 import java.util.Date;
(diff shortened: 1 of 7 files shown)


schildbach:
I took the liberty and split your first commit into two (the refactoring and adding the TRIPS_VIA cap).

About the REALTIME caps, I'm not really confident they're useful. It depends on individual vehicels (e.g. bus) wether realtime data is available or not.

ialokim:
Thanks a lot for merging the `TRIPS_VIA` capability!

---

> About the REALTIME caps, I'm not really confident they're useful. It depends on individual vehicels (e.g. bus) wether realtime data is available or not.

As already stated in https://github.com/schildbach/public-transport-enabler/pull/210#issuecomment-462170130, I think it would nevertheless make sense for e.g. hiding the refresh functionality altogether in a client, if the API does never expose realtime data (as e.g. Navitia based providers).

Just relying on the actual availability of realtime data in a specific trip at a specific point in time might not be enough: It might happen that a provider first doesn't send realtime information (e.g. if the vehicle is still not registered in the system or the trip lies some hours in the future) but later would be able to serve realtime data. A `Trip.hasRealtime()` method would then not be useful for the decision whether the refresh button should be showed or hidden.

PR #364, opened 2021-04-03
VRS: parse unknown from

alexander-albers:
VRS may return `{"error":"Fehlerhafter Start"}` when querying trips.


schildbach:
Merged.

PR #402, opened 2021-09-17
AbstractNavitiaProvider: Remove all providers based on Navitia.

schildbach:
I propose to remove all providers based on Navitia. They have been causing many runtime errors since years, without anyone working on them. If we don't find a responsible maintainer for the Navitia API (mainly AbstractNavitiaProvider), we I think we'd rather remove these classes.


schildbach:
Pinging @aelkhour and @grote as you're listed as authors.

grote:
Also pinging @ialokim and @prhod about the removal of Navitia from PTE.

ialokim:
I think it's less the PTE's implementation of `AbstractNavitiaProvider` that leads to problems, but instead the data on the Navitia server which is not always up-to-date or gets some updates changing IDs or such. I think we can not do much on PTE's side on this, apart from perhaps implementing a slightly better error handling.

It would be really a shame if all those providers were removed from PTE. I would be okay maintaining them to the extend that is possible without having access to the data source (Navitia API) itself.

aelkhour:
@schildbach I don't have the bandwidth to maintain this provider. Although I am the instigator of this and had a lot of fun adding it, I understand your decision if there are no maintainers.

schildbach:
> The removal of all providers based on Navitia would be a big loss :( What specific errors/problems are encountered with Navitia? I would be happy to help working those out :)

Sorry for getting back to your question only now.

I tagged most of the open issues with the tag "navitia".

schildbach:
There is one more problem with Navitia: their API limit for free accounts seems to be exceeded by Öffi now and then. Not sure how other apps like Transportr do, but for Öffi this is quite discouraging.

schildbach:
Merged.

Issue #401, opened 2021-09-09
VRS: nearby stations does not work anymore


alexander-albers:
VRS made some changes and now requests to `tx_vrsinfo_ass2_timetable` return `unknown error: Keine gültige ID` for coord requests. This means using this method for querying nearby departures does not work anymore.

~I think I found a solution to this problem~ the https endpoint also does not seem to work for this type of request.

alexander-albers:
The correct solution is to use http://android.vrsinfo.de/index.php?eID=tx_ekap_here&ta=vrs&&lat=&lon= instead.

Hope this helps.

PR #407, opened 2021-09-27
VRS: Fix queryNearbyLocations

mdyrna:
https://github.com/schildbach/public-transport-enabler/issues/401

### DIFF
--- a/src/de/schildbach/pte/VrsProvider.java
+++ b/src/de/schildbach/pte/VrsProvider.java
@@ -25,14 +25,11 @@ import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Calendar;
-import java.util.Collections;
-import java.util.Comparator;
 import java.util.Currency;
 import java.util.Date;
 import java.util.EnumSet;
 import java.util.GregorianCalendar;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
@@ -47,9 +44,6 @@ import javax.annotation.Nullable;
 import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import com.google.common.base.Strings;
 
 import de.schildbach.pte.dto.Departure;
@@ -89,8 +83,6 @@ public class VrsProvider extends AbstractNetworkProvider {
             Capability.TRIPS,
             Capability.TRIPS_VIA
     );
-
-    private static final Logger log = LoggerFactory.getLogger(VrsProvider.class);
 
     @SuppressWarnings("serial")
     private static class Context implements QueryTripsContext {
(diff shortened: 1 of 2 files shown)


src/de/schildbach/pte/VrsProvider.java
```diff
@@ -718,6 +659,9 @@ else if (error.equals("Server Error"))
                 else if (error.equals("Keine Verbindungen gefunden."))
                     return new QueryTripsResult(new ResultHeader(NetworkId.VRS, SERVER_PRODUCT),
                             QueryTripsResult.Status.NO_TRIPS);
+                else if (error.contains("keine") && error.contains("Verbindungen"))
```
schildbach, line 662: This match feels too broad to me. Can we not just check for the specific message, like the others?

test/de/schildbach/pte/live/VrsProviderLiveTest.java
```diff
@@ -121,14 +120,16 @@ public void nearbyStationsWithLimits() throws Exception {
...
+                Location.coord(50732100, 7096820), 0, 1);
         print(result2);
 
         final NearbyLocationsResult result3 = queryNearbyLocations(EnumSet.of(LocationType.STATION),
-                Location.coord(50732100, 7096820), 1000, 50);
+                Location.coord(50732100, 7096820), 100, 0);
         print(result3);
     }
 
+    @Ignore
```
schildbach, line 131: Es the error 500 happening because of the parameters of our query (both 0), or because the result set would be empty?

In the second case I think we should handle this in the provider. HTTP status codes is not something the apps want to deal with.

mdyrna: The Java method arguments types, maxDistance and maxLocations are not passed to the API, but are used for internal filtering. (When I have more time, I might reverse engineer the VRS app to find out if the API supports these parameters.)

The resulting URL for the test case is http://android.vrsinfo.de/index.php?eID=tx_ekap_here&ta=vrs&lat=0.000001&lon=0.000000 (a location far outside the VRS area).

The result body is:
```
{"fehler":"Ung\u00fcltige Umkreis-Koordinate"}
```

But unfortunately, the status code is 500 so that de.schildbach.pte.util.HttpClient.getInputStream throws an InternalErrorException, which cannot be handled gracefully by the provider implementation.

src/de/schildbach/pte/VrsProvider.java
mdyrna: The literal error message was:

>>>
Es wurden keine gültigen Verbindungen für diese Anfrage gefunden.
>>>

I was hesitant because of the u Umlaut in "gültigen", but not what I read through the already implemented error messages, please feel free to replace the if case with the literal message quoted above.

test/de/schildbach/pte/live/VrsProviderLiveTest.java
mdyrna: I just checked that if both coordinates are != 0, a valid result set is returned.

So if you replace
```
Location.coord(1, 0)
```
with
```
Location.coord(1, 1)
```

the API returns four results from Switzerland: http://android.vrsinfo.de/index.php?eID=tx_ekap_here&ta=vrs&lat=0.000001&lon=0.000001

So please feel free to replace the Location.coord arguments, re-activate the test case and assert 4 results.

mdyrna: ...and if we now set maxDistance to 1000 m instead of 0 (unlimited), the original purpose of the test case is restored:

```
    @Test
    public void nearbyLocationsEmpty() throws Exception {
        final NearbyLocationsResult result = queryNearbyLocations(EnumSet.allOf(LocationType.class),
                Location.coord(1, 1), 1000, 0);
        print(result);
        assertEquals(0, result.locations.size());
    }
```

mdyrna:
I committed the two discusses small changes. @schildbach Can you please continue merging?

src/de/schildbach/pte/VrsProvider.java
schildbach: I think if there is a problem with Umlauts it would be a good opportunity to check/fix the encodings.

schildbach:
Squashed & merged. Thanks!

PR #414, opened 2021-10-23
switch MVV to new endpoint with realtime data

johan12345:
For MVV München, the EFA endpoint that the PTE is currently using (https://efa.mvv-muenchen.de/mobile/) does not provide realtime data (not sure if it did that in the past, but right now it does not).

There are two new endpoints that are currently in use on the MVV website and which have realtime data:
- https://m.mvv-muenchen.de/jqm/mvv_lite (used by the [mobile website](https://m.mvv-muenchen.de/mvvMobile5/de/))
- https://efa.mvv-muenchen.de/ng/ (used by the [desktop website](https://efa.mvv-muenchen.de/index.html))

I couldn't get the former one to work with PTE as it seems to use a different XML format (pretty similar to the `...Mobile` functions, but not the same), but the latter one works well, so I implemented that one in this PR.

**Notes:**
- The station IDs have changed in this new endpoint - IDs that were previously 1-4 digit numbers have received an additional prefix `91000...` (e.g. München Marienplatz is now `91000002` instead of `2`), and longer IDs that already started with `1000...` also got the additional `9` in front. Is it necessary to add some compatibility code to convert the old station IDs into the new ones?
- Tests that include a `queryMoreTrips` call currently fail with a `SessionExpiredException`. I am not yet sure why that is, the session cookie and `requestID/sessionID` parameters seem to be passed correctly.

### DIFF
--- a/src/de/schildbach/pte/MvvProvider.java
+++ b/src/de/schildbach/pte/MvvProvider.java
@@ -39,7 +39,7 @@ import okhttp3.HttpUrl;
  * @author Andreas Schildbach
  */
 public class MvvProvider extends AbstractEfaProvider {
-    private static final HttpUrl API_BASE = HttpUrl.parse("https://efa.mvv-muenchen.de/mobile/");
+    private static final HttpUrl API_BASE = HttpUrl.parse("https://efa.mvv-muenchen.de/ng/");
 
     public MvvProvider() {
         this(API_BASE);
@@ -50,7 +50,7 @@ public class MvvProvider extends AbstractEfaProvider {
         setIncludeRegionId(false);
         setRequestUrlEncoding(Charsets.UTF_8);
         setStyles(STYLES);
-        setSessionCookieName("SIDefaalt"); // SIDefa
+        setSessionCookieName("SIDefa");
     }
 
     @Override
--- a/src/de/schildbach/pte/MvvProvider.java
+++ b/src/de/schildbach/pte/MvvProvider.java
@@ -17,7 +17,12 @@
 
 package de.schildbach.pte;
 
+import java.io.IOException;
+import java.util.Calendar;
+import java.util.GregorianCalendar;
+import java.util.Date;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -27,12 +32,17 @@ import javax.annotation.Nullable;
 import com.google.common.base.Charsets;
 
 import de.schildbach.pte.dto.Line;
+import de.schildbach.pte.dto.Location;
 import de.schildbach.pte.dto.Point;
 import de.schildbach.pte.dto.Position;
 import de.schildbach.pte.dto.Product;
+import de.schildbach.pte.dto.QueryTripsContext;
+import de.schildbach.pte.dto.QueryTripsResult;
 import de.schildbach.pte.dto.Style;
 import de.schildbach.pte.dto.Style.Shape;
 
+import de.schildbach.pte.dto.Trip;
+import de.schildbach.pte.dto.TripOptions;
 import okhttp3.HttpUrl;
 
 /**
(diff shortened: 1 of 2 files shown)


johan12345:
Hi @schildbach, I have implemented a workaround to make queryMoreTrips work with the new API despite the load balancer in 72ef473. Instead of using the session-based `tripNext` command, it simply starts a new query based on the departure time of the last trip in the previous result (and similarly for previous trips). The same algorithm is used on the [mobile version](https://m.mvv-muenchen.de/mvvMobile5/de/index.html#trips) of the MVV website, so the results should be the same.

Do you think this is a viable solution? As I said, the same measures would also be required when implementing the "new mobile API", as it uses the same load balancer.

alexander-albers:
> I have implemented a workaround to make queryMoreTrips work with the new API despite the load balancer in https://github.com/schildbach/public-transport-enabler/commit/72ef473998cbf8d808fcf013759e9fe655ae3eef. Instead of using the session-based tripNext command, it simply starts a new query based on the departure time of the last trip in the previous result (and similarly for previous trips). The same algorithm is used on the [mobile version](https://m.mvv-muenchen.de/mvvMobile5/de/index.html#trips) of the MVV website, so the results should be comparable.
Do you think this is a viable solution? As I said, the same measures would also be required when implementing the "new mobile API", as it uses the same load balancer.

FYI, I've had the same issue also with VRN provider in [TripKit](https://github.com/alexander-albers/tripkit) (this probably applies to pte as well). VGN seems to use the same approach on their website as MVV, as far as I can tell. I've implemented this workaround (https://github.com/schildbach/public-transport-enabler/commit/72ef473998cbf8d808fcf013759e9fe655ae3eef) in [my Swift project](https://github.com/alexander-albers/tripkit/commit/fe8cc37d962b926ad859a4fea57ef06768362a15) and could not find any instance where there is a mismatch in trips between the old session based version and your proposed stateless version.

johan12345:
@alexander-albers Thanks, this looks good!

@schildbach, have you had time to take look at the updated PR so far? If you want, I could also implement this workaround as an optional setting for the `AbstractEfaProvider` as @alexander-albers did it in his project, instead of a specific change in the `MvvProvider` subclass, so that the change can be reused for VRN.

shilch:
Hi, any update on this? It would be really helpful to get realtime data for MVV because for S-Bahn delays I have to consult the Deutsche Bahn inside Öffi. Also, several times I had the situation where Öffi didn't show a good connection that was, however, shown in the MVV mobile app. I guess the new API does a better job.

schildbach:
Is the workaround regarding the loadbalancer still needed? I admit that don't really like such workarounds – I vaguely remember doing something similar in the past and it wasn't relyable either.

johan12345:
Yes, unfortunately it still seems to be the same - if I remove the workaround and run the `shortTrip()` test in `MvvProviderLiveTest` many times, it only succeeds about 10% of the time (-> only if we happen to reach the same server again).

As I said, the [mobile version](https://m.mvv-muenchen.de/mvvMobile5/de/index.html#trips) of the MVV website uses the exact same workaround (looking at the [minified JS](https://m.mvv-muenchen.de/mvvMobile5/202307280701/js/mvv.min.js), functions `MdvLiteTrips.prototype.get_next_trip`/`.get_prev_trip` and `.get_request_datetime`).

For the [desktop version](https://efa.mvv-muenchen.de/), we don't know exactly how it does it, because the next/previous trip request is done not directly to the standard XML or JSON EFA endpoints, but through a separate `https://efa.mvv-muenchen.de/xhr_trips` endpoint whose output is HTML.

src/de/schildbach/pte/MvvProvider.java
```diff
@@ -157,4 +166,117 @@ protected Position parsePosition(final String position) {
...
+
+        if (requestingMoreTrips) {
+            // ensure that the first displayed trip is after the given departure time /
+            // last displayed trip is before the given arrival time
+            url.addEncodedQueryParameter("calcOneDirection", "1");
+        }
+    }
+
+    private Date addMinutesToDate(Date initial, int minutes) {
+        Calendar c = Calendar.getInstance();
```
schildbach, line 277: Maybe  `Calendar c = new GregorianCalendar(timeZone);`? The time zone for MvvProvider is CET.

johan12345: Fixed ✅

schildbach:
I just merged your first commit, the API base change. I'll review the session workaround separately.

schildbach:
In Öffi, I plan to migrate all locally persisted MVV station IDs (favorites, saved trip queries) like this:

* [0…10000[ → add 91000000
* [10000…1000000[ → add 90000000

Does this sound reasonable?

schildbach:
Quoting from the PR description:

> Is it necessary to add some compatibility code to convert the old station IDs into the new ones?

No, not in PTE as it is stateless. If you keep state in your app (e.g. favorites), you need to add migration code there.

schildbach:
> * [1000000…2000000[ → add 90000000

I think I'll do away with this migration. The 1000000…2000000 ID range is still used, even on the new endpoint (e.g. Starnberg).

johan12345:
Yes, true. Let's just migrate the [0…10000[ range (mostly stations within Munich)

Issue #502, opened 2022-12-14
SNCB: java.lang.RuntimeException: CLIENTVERSION 


gargamel17:
### Query
- NetworkId: `SNCB`
- From: `Location{ADDRESS, 50.6966880/5.4057460, place=Crisnée, name=Rue Vincent Bonnechère BE-WAL, BE-WLG}`
- Via: `null`
- To: `Location{STATION, 262348, 50.6948550/5.2500870, name=WAREMME Gare [TEC], products=[HIGH_SPEED_TRAIN, REGIONAL_TRAIN, BUS]}`
- Date: `Wed Dec 14 11:40:09 GMT+01:00 2022`
- Departure: `true`
- Products: `[HIGH_SPEED_TRAIN, REGIONAL_TRAIN, SUBURBAN_TRAIN, SUBWAY, TRAM, BUS, FERRY, CABLECAR, ON_DEMAND]`
- Optimize for: `LEAST_DURATION`
- Walk Speed: `NORMAL`

### Error
```
java.lang.RuntimeException: CLIENTVERSION 
de.schildbach.pte.AbstractHafasClientInterfaceProvider.jsonLocMatch(AbstractHafasClientInterfaceProvider.java:475)
de.schildbach.pte.AbstractHafasClientInterfaceProvider.jsonTripSearchIdentify(AbstractHafasClientInterfaceProvider.java:523)
de.schildbach.pte.AbstractHafasClientInterfaceProvider.jsonTripSearch(AbstractHafasClientInterfaceProvider.java:544)
```

### Additional information
[Please modify this part]

Muspah:
SNCB deprecated their HAFAS endpoint and switched to an Azure-hosted HAFAS OpenAPI endpoint with custom arrival/departure boards.

schildbach:
What's "HAFAS OpenAPI"? Do you have some link?

Muspah:
An older version of the API spec can be found in: https://projects.shift2rail.org/download.aspx?id=4f1ad68b-0e62-4183-98ae-7232a5c519bd

SNCB only offers JSON endpoints and only specific endpoints are available. TrainSearch is missing for example, looking up a specific journey requires multiple calls and/or caching JourneyRef's.

schildbach:
Just for reference, the highest version the endpoint knows about is `1.51` – yields `CLIENTVERSION`. Any higher version yields `VERSION Invalid version`.

schildbach:
Also see:
[...]

PR #516, opened 2023-02-11
SNCB: remove provider for Belgium

schildbach:
As far as I can tell, their HCI API is not working any more. It yields the `CLIENTVERSION` error for each request, and increasing the client version doesn't change this (up to a point where the version itself is invalid). Their website and apps have been migrated to use a different API (partly "open", partly proprietary) which we currently do not support.

I think it's up to the "Nationale Maatschappij der Belgische Spoorwegen" to revive their HCI – otherwise we should remove support on our side. We can implement their new API later, but someone needs to volunteer.

Also see #502.


Muspah:
The API is deprecated and being shutdown, the response (used to) contain(s):

Cette fonctionnalité n'est plus disponible. Pour planifier votre voyage, téléchargez la nouvelle app SNCB.
Deze app is niet langer beschikbaar. Om jouw reis te plannen download je best de nieuwe NMBS-app.
This feature is no longer available. To plan your trip, download the new SNCB app.
Diese App wird nicht mehr unterstützt. Für die Planung Ihrer Reise laden sie bitte die neue SNCB App herunter.

schildbach:
Merged.

PR #533, opened 2023-05-22
Update Hagen line colours

metrophil:
updated Hagen line colours according to upcoming timetable change on 11 June 2023


schildbach:
Merged, thanks!

PR #541, opened 2023-08-21
Added an IsraelProvider

mikwee:
I admit I have not much knowledge in the topic, this is my first-ever pull request in fact. I really do hope the Navitio API has Israel support, because they do not have info on that on their website(s). But the bottom line is, I want Transportr to work in Israel too, and I realize this is a major part of that road.
Thank you for your patience.


schildbach:

PR #554, opened 2024-01-16
AVV_Aachen: handle inconsistent place suffix for some stations

thtomate:
Some stations in Aachen have entries with the three differtent names "station name, AC", "Aachen, station name" and "AC, station name" in AVV's Hafas. Some others have "station name, Aachen" and "Aachen, station name". If you type the station name first, you get the variant "station name, AC" resp. "station name, Aachen" . Without the filter proposed here, the station name is parsed as "AC" (with place "station name"), which is not helpful in any abbreviating views, especially with multiple popular stations having this issue.
In Öffi for example when selecting such an entry for Bushof and Turmstraße: photo_2024-01-16_19-41-17

Meanwhile, for the route results, the API returns station names formatted "Aachen, station name" which looks fine.

This PR implements a filter for handling the station names ~~ending with ", AC" directly~~with "AC", ending in "Aachen" and a test checking the station name showing up at the correct place.

### DIFF
--- a/src/de/schildbach/pte/AvvAachenProvider.java
+++ b/src/de/schildbach/pte/AvvAachenProvider.java
@@ -46,11 +46,26 @@ public class AvvAachenProvider extends AbstractHafasClientInterfaceProvider {
         setApiAuthorization(apiAuthorization);
     }
 
+    private static final String[] PLACES = { "AC", "Aachen" };
+
     @Override
     protected String[] splitStationName(final String name) {
+        // Some stations in Aachen city have 3 names: "Aachen, station name", "station name, AC" and "AC, station name".
+        // Some (other) stations has 2 variants: "Aachen, station name" and "station name, Aachen"
+        // If you type the station name first, you get the variant "station name, AC" resp. "station name, Aachen",
+        // which would be parsed as a station AC resp. Aachen in "station name".
+        for (final String place: PLACES) {
+            if (name.endsWith(", " + place)) {
+                return new String[] { "Aachen" , name.substring(0, name.length() - place.length() - 2) };
+            }
+        }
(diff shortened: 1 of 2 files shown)


schildbach:
> Some stations in Aachen have the names "station name, AC", "Aachen, station name" and "AC, station name" in AVV's Hafas. 

Do you know if those are 3 different station IDs, or does Hafas somehow know alternative names for the same station ID?

src/de/schildbach/pte/AvvAachenProvider.java
```diff
@@ -48,6 +48,12 @@ public AvvAachenProvider(final String apiClient, final String apiAuthorization)
 
     @Override
     protected String[] splitStationName(final String name) {
+        // Some stations in Aachen city have both the names "station name, AC" and "AC, station name".
+        // If you type the station name first, you get the variant "station name, AC", which would be parsed as the
+        // station AC in "station name".
+        if (name.endsWith(", AC")) {
+            return new String[] { "Aachen" , name.substring(0, name.length() - ", AC".length()) };
+        }
```
schildbach, line 56: I'm not sure, but shouldn't both of the above variants be split into "Aachen" and the plain station name?

thtomate: I don't care (They are already split into "AC" and plain station name).

thtomate:
> Do you know if those are 3 different station IDs

No. They have the same ID

src/de/schildbach/pte/AvvAachenProvider.java
thtomate: You added another filter for that

thtomate:
There are also other variations. I will amend the fix the next days

schildbach:
Merged, thanks!

PR #571, opened 2024-04-29
MVV: add new Munich S-Bahn line S5

infanf:
Add new line that will run from December 2024
https://www.s-bahn-muenchen.de/fahren/s5


schildbach:
Thanks for the contribution! Merged.

PR #590, opened 2024-08-13
Fix MOT number and colours for Glasgow subway

NobleTwo:
At the moment, the public-transport-enabler doesn't recognise the Glasgow subway (SPT) as a subway. This MR fixes that. While at it, it also changes the colours of the two subways lines to those used in the maps.

Official maps:

Before:

before

After:
after


NobleTwo:
Is there any chance of getting this reviewed or merged, @schildbach? (Apologies if I overlooked any contributing guidelines!)

Issue #610, opened 2024-12-06
DB: SERVICE_DOWN


Nibunnoichi:
### Query
- NetworkId: `DB`
- From: `Location{STATION, 8006713, 51.0314560/6.9915040, name=Leverkusen Mitte, products=[REGIONAL_TRAIN, SUBURBAN_TRAIN, BUS, ON_DEMAND]}`
- Via: `null`
- To: `Location{STATION, 443410, 50.9202960/7.0819450, place=Köln, name=Rath/Heumar, products=[TRAM]}`
- Date: `Fri Dec 06 12:07:01 GMT+01:00 2024`
- Departure: `true`
- Products: `[HIGH_SPEED_TRAIN, REGIONAL_TRAIN, SUBURBAN_TRAIN, SUBWAY, TRAM, BUS, FERRY, CABLECAR, ON_DEMAND]`
- Optimize for: `LEAST_DURATION`
- Walk Speed: `NORMAL`

### Error
```
Die Datenquelle ist momentan nicht erreichbar :(
Bitte versuche es später noch einmal.
```

### Additional information
[Please modify this part]

traines-source:
The legacy DB HAFAS API seems to become increasingly unreliable, returning errors. For more context, see https://github.com/public-transport/hafas-client/issues/331#issuecomment-2529430123

schildbach:
Thanks for the head up! Do you have some source of the EOL date?

I won't have time to implement a new API for at least until end of February. Also, I'm reluctant to invest into implementing a proprietary API (one that is not an "industry standard" like HCI or EFA). I'd rather work on implementing Transitous or similar.

That said, of course the loss of DB data is a pretty big issue. So I'd welcome contributions towards "Vendo" into PTE.

traines-source:
"Someone" on the [OKFN transport channel](https://matrix.to/#/!fGsJOaXhgnTCFCghEl:matrix.org/$Af-pOVmexMHJ9b5YYtV2ThMl5loSI0oY9YaPhdP3GNU?via=matrix.org&via=kde.org&via=tchncs.de) said that they apparently want to shut it off after the timetable change of 14/15th December. And from my point of view many things have been indicating it, e.g. the missing price information after the timetable change and the brownout (?) of recent weeks.

[...]

PR #614, opened 2024-12-20
Add DB-Movas Provider

traines-source:
As discussed in https://github.com/schildbach/public-transport-enabler/issues/610, I have now implemented the DB Vendo/Movas API.
The tests are green and I have manually tested with Oeffi, everything looks fine, but more testing surely can't harm.
Some notes:
* The nearby endpoint doesn't allow for POI search, but that doesn't seem to be surfaced in Oeffi anyways
* The departure boards are limited to one hour, but I think this is OK
* More messages/remarks/notices seem to be shown than before, although I'm only surfacing the important ones (usually disruptions, but also static ones like phone numbers for on demand buses etc. With the legacy DB Provider, some "Rufbus" were not distinguishable from normal buses, e.g. from Rothenburg o.d.T. to Dombühl at 20:31).
* The cheapest available adult price is shown.
* Error handling is not (yet) very granular

I noticed that Oeffi always gulps the last departure on the departure boards, because in [StationDetailsActivity.java:438](https://gitlab.com/oeffi/oeffi/-/blob/main/oeffi/src/de/schildbach/oeffi/stations/StationDetailsActivity.java?ref_type=heads#L438) it should be `selectedDepartures.size()+1` due to position 0 being a dummy position. This becomes much more obvious due to the one hour limitation of the new API (and at this time of the day :).

Let me know if you also want a PR for Oeffi, however I've only quickly added the new provider and one would need to decide whether one wants to completely replace the old DB Provider (ironically, right now, the HAFAS API seems to work beautifully again, let's see for how long...)


schildbach:
Huge thanks for this PR!

> The nearby endpoint doesn't allow for POI search, but that doesn't seem to be surfaced in Oeffi anyways

That's right. "Öffi Stations" is only about stations/stops, not POIs or street addresses.

> The departure boards are limited to one hour, but I think this is OK

Is that a limit from the API? I think 1 hour is very short, given that delays can easily be more than one hour. But of course it's better than nothing. Maybe in future we can ask DB to extend it.

> More messages/remarks/notices seem to be shown than before, although I'm only surfacing the important ones

The old messages/categories were badly curated (by me, ages ago), so feel free to apply whatever filter feels right. We don't need to copy old behaviour here.

> The cheapest available adult price is shown.

Cool! We never had prices with DB, afair.

> Error handling is not (yet) very granular

Yes, that's expected. There is never any API doc about errors, so we need to fine-tune them over the coming weeks and months.

> Let me know if you also want a PR for Oeffi

Sure, feel free! But I can also look after this bug if you prefer.

> one would need to decide whether one wants to completely replace the old DB Provider

This is a very important decision indeed. From my gut feeling, I'd like to stay with the old provider as long as it works reasonably reliable. The HCI implementation has proven to be pretty solid over the years.

We could maybe provide the Movas API as an alpha alongside HCI, to help testing? If the HCI is switched off permanently, we can migrate everything to just one "DB" network ID. Related: Do you know if stations IDs have changed between the APIs? If so, have you witnessed any pattern, like adding a fixed constant?

I should have brought this up earlier, it's maybe too late now: Currently the PTE uses the old (and probably deprecated?) Android JSON API (org.json.JSONObject). At some point – and we could have started with this impl – I think it would make sense to migrate to something more modern. I always had Moshi in mind, because it is built on OkIo. But I'm not sure if it's still the right choice – we also have access to almost all of JDK 8 API nowadays, and in maybe 2 years JDK 11+.

traines-source:
> Is that a limit from the API? I think 1 hour is very short

At least I haven't yet found out a parameter with which to influence it. Once could of course do multiple requests under certain conditions (such as not enough departures returned), but there is not even a `laterRef` or something, DB Navigator just requests for one hour later if you click on "later" :see_no_evil: 

> Maybe in future we can ask DB to extend it.

Maybe you have better contacts into DB, from my experience they won't do something like that. On a side note, I haven't asked for permission so far to use the API, but at least it's without any API Key this time. Or one could switch to [another DB API](https://github.com/public-transport/db-vendo-client) for the boards that support longer durations, but which have other drawbacks:

* RIS::Stations API, needs an API Key, a free one having a very low quota
* bahnhof.de API, which only allows querying for the current datetime (which would be fine for Oeffi I guess) 
* regio-guide.de API, which doesn't surface cancelled trips (which is currently also the behavior of Oeffi/DB Provider) and no messages/notices either

Which is why already for the sake of consistency I had opted for the API that DB Navigator uses for both routing and boards. And I personally would propose to show cancelled trips in the departure board, even if there is no cancelled flag in PTE for them, at least they now reliably show "Halt entfällt"/"Stop cancelled".

> We could maybe provide the Movas API as an alpha alongside HCI, to help testing?

Sounds good, I'll make a PR for Oeffi then since that's basically what I've done.

> Do you know if stations IDs have changed between the APIs?

It still uses the EVA numbers and in the queries the `lid` created out of that. But for POI and ADDRESS, PTE before used the complete `lid` as an id as far as I can tell. I can change it to also use that, then the serialized locations should be compatible I guess.

> Android JSON API (org.json.JSONObject)

Indeed I've used it throughout and since it does not seem to be deprecated and it works nicely, I am a bit reluctant to change everything now :D But if you want I can have a look.

src/de/schildbach/pte/NetworkId.java
```diff
@@ -25,7 +25,7 @@ public enum NetworkId {
     RT,
 
     // Germany
-    DB, BVG, VBB, NVV, BAYERN, MVV, INVG, AVV_AUGSBURG, VGN, VVM, VMV, SH, GVH, BSVAG, VBN, NASA, VMT, VVO, VGS, VRR, VRS, AVV_AACHEN, MVG, VRN, VVS, DING, KVV, NVBW, VVV,
+    DB, DBMOVAS, BVG, VBB, NVV, BAYERN, MVV, INVG, AVV_AUGSBURG, VGN, VVM, VMV, SH, GVH, BSVAG, VBN, NASA, VMT, VVO, VGS, VRR, VRS, AVV_AACHEN, MVG, VRN, VVS, DING, KVV, NVBW, VVV,
```
schildbach, line 28: Given the fact that the HCI API now seems to be shut off, I'm now leaning towards replacing the old ID. If on the other hand there is a chance the old API will be ressurected, I'd keep it separate. What do you think?

Sorry for the back and forth about this one.

schildbach (commented):
I think this looks really good overall.

src/de/schildbach/pte/util/HttpClient.java
```diff
@@ -236,9 +236,12 @@ public void getInputStream(final Callback callback, final HttpUrl url, final Str
...
         request.headers(Headers.of(headers));
-        if (postRequest != null)
-            request.post(RequestBody.create(MediaType.parse(requestContentType), postRequest));
-        request.header("Accept", SCRAPE_ACCEPT);
+        if (postRequest != null) {
+            final MediaType m = requestContentType != null ? MediaType.parse(requestContentType) : null;
+            request.post(RequestBody.create(m, postRequest));          
+        }
+        if (!headers.containsKey("Accept"))
+            request.header("Accept", SCRAPE_ACCEPT);
```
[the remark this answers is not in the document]
traines-source: Separated it into two commits now, but kept the logic as is. And it's really just to make it possible to pass in custom Accept and Content-Type headers. The `MediaType` logic would add a `charset=utf-8` suffix, which the DB API would not accept, because they use the Content-Type for API versioning and want it to match exactly apparently.

src/de/schildbach/pte/DbMovasProvider.java
```diff
@@ -0,0 +1,799 @@
...
+        this.departureEndpoint = API_BASE.newBuilder().addPathSegments("bahnhofstafel/abfahrt").build();
+        this.tripEndpoint = API_BASE.newBuilder().addPathSegments("angebote/fahrplan").build();
+        this.locationsEndpoint = API_BASE.newBuilder().addPathSegments("location/search").build();
+        this.nearbyEndpoint = API_BASE.newBuilder().addPathSegments("location/nearby").build();
+        this.resultHeader = new ResultHeader(network, "DB (movas)");
+    }
+
+    private String doRequest(final HttpUrl url, final String body, final String contentType) throws IOException {
+        httpClient.setHeader("X-Correlation-ID", "null");
+        httpClient.setHeader("Accept", contentType);
```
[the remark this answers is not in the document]
traines-source: Yes, without them exactly as they are (ofc DB Navigator sets the correlation id to some UUID for tracking, but apparently any string works), it returns HTTP 400 or 500 or sth similar.

src/de/schildbach/pte/NetworkId.java
traines-source: I don't think that the API will return because 1) many things have hinted at it being turned off 2) people from within DB have said that this would probably happen 3) it just makes sense for them to turn it off, and at this point in time after coming back from the christmas holidays. But ofc I might be wrong.
Let me know or just change it yourself. Anyways it's probably a good idea if you quickly test the forward compatibility of the serialized locations and journeys etc from Db to DbMovas. I can't do that anymore, since I had to uninstall due to signing and have barely anything in my Oeffi search history for the old DB Provider and for obvious reasons can't create any new entries now...

traines-source: You could also keep the old one around as `DB_HAFAS` or sth and mark it as deprecated in the Oeffi UI, so that in the worst case one can still select it, albeit without saved journeys.

derhuerst: Newbie question: Does labelling the new client as `DB` have any benefits?

alangecker: Would reusing the old ID avoid the need for users to switch their network in the settings?

In this case this would be, in my opinion, a strong reason to do it, so that for users after an update everything works again as they are used to without the need for an manual procedure.

derhuerst: >> Would reusing the old ID avoid the need for users to switch their network in the settings?
> 
> In this case this would be, in my opinion, a strong reason to do it, so that for users after an update everything works again as they are used to without the need for an manual procedure.

I'd consider that a responsibility of the Öffi code, not `public-transport-enabler`'s, but maybe building that into Öffi is not worth the complexity.

schildbach: Both reasonings are valid I'd say. In the past, if an API has a designated successor *and* IDs are compatible I (more or less transparently) replaced it at PTE level.

schildbach: > You could also keep the old one around as `DB_HAFAS` or sth and mark it as deprecated in the Oeffi UI, so that in the worst case one can still select it, albeit without saved journeys.

I'd rather not do that, as it seems to accumulate all the downsides: needs migration code in Öffi, adds another (sooner or later redundant) ID. To ressurect the old API if needed, git would be our friend I guess.

src/de/schildbach/pte/util/HttpClient.java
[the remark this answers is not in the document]
schildbach: I've just come up with a small alternative PR to allow overriding the Accept header:
https://github.com/schildbach/public-transport-enabler/pull/619
Would you like to have a look at it?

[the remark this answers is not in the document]
schildbach: I just merged your fix to the handling of the `requestContentType` argument as bb11f15d3c32584c0e6d54200a9ff7e2af68a968. Seems this bug was there for a longer time already and broke several `getInputStream()` and `get()` variants.

src/de/schildbach/pte/DbMovasProvider.java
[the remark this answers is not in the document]
schildbach: Ok, good to know. I think we should move the comment below actually *above* the three `setHeader` statements and maybe reword it a bit so it's valid for setting custom headers in general.

Or, at least move it one line up because we generally place comments *before* the line to comment. This is just a nit-pick, of course.

src/de/schildbach/pte/util/HttpClient.java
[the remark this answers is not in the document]
traines-source: Ah didn't even realize that `null` was passed somewhere before, but yes. I suppose you rebase/cherry pick/whatever this PR then as soon as you've merged the other one, or shall I do that?

src/de/schildbach/pte/DbMovasProvider.java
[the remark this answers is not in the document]
traines-source: I placed it above this line because I'm passing `null` to `httpClient.get` instead of the content-type as in most other Provider impls, precisely because it would add the charset. But can ofc move it up and elaborate

schildbach:
I just tested this against Öffi, and it looks very good. Sadly I didn't have addresses or POIs in my history, but stations are perfectly compatible.

src/de/schildbach/pte/DbMovasProvider.java
```diff
@@ -0,0 +1,799 @@
...
+    private static final Pattern P_SPLIT_NAME_FIRST_COMMA = Pattern.compile("([^,]*), (.*)");
+    private static final Pattern P_SPLIT_NAME_ONE_COMMA = Pattern.compile("([^,]*), ([^,]*)");
+
+    public DbMovasProvider() {
+        super(NetworkId.DBMOVAS);
+        this.departureEndpoint = API_BASE.newBuilder().addPathSegments("bahnhofstafel/abfahrt").build();
+        this.tripEndpoint = API_BASE.newBuilder().addPathSegments("angebote/fahrplan").build();
+        this.locationsEndpoint = API_BASE.newBuilder().addPathSegments("location/search").build();
+        this.nearbyEndpoint = API_BASE.newBuilder().addPathSegments("location/nearby").build();
+        this.resultHeader = new ResultHeader(network, "DB (movas)");
```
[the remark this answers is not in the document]
traines-source: I think previously something like "provided by Deutsche Bahn" was shown, but that happened somewhere in Oeffi   (or now, if you have replaced `DB`, it shows as well for the new Provider).

traines-source:
Well, I guess we'll have to hope for the best with addresses then (but I think it should work, and it's a nice to have, not critical anyways IMHO).
I suppose you have then done the rebasing and will do the amendments to `ResultHeader` and comments regarding request headers yourself, otherwise lmk.

src/de/schildbach/pte/NetworkId.java
traines-source: The diff between the two tests shows that some changes are necessary,
1) different constructor
2) unsupported POI search
3) station `692990` that only very rarely has departures and thus fails frequently due to the 1 hour limit (on that note, don't forget the `selectedDepartures.size()+1` fix in Oeffi, it's important for stops where there is one departure per hour)
4) changed id of München - Berg am Laim (don't recall exactly)
5) fix some evergreens/fail early with additional asserts
6) an additonal test checking querying for stations only by name and with backwards compatible location address/poi ids
6) some changed errors, because for too close stations the new API just returns an empty result
So rather keep the new Test or feel free to check everything again if you want :)

schildbach:
Feel free to rebase and squash, and add any last amendments.On Jan 10, 2025 16:05, Traines ***@***.***> wrote:
Well, I guess we'll have to hope for the best with addresses then (but I think it should work, and it's a nice to have, not critical anyways IMHO).
I suppose you have then done the rebasing and will do the amendments to ResultHeader and comments regarding request headers yourself, otherwise lmk.

traines-source:
Nice 👍 However, as far as I can see, you haven't fixed the `selectedDepartures.size()+1` issue. As I said I think this is important as there are many stops where there is exactly one departure per hour, which will then not be shown (and even more confusing for the user, shown in the nearby view, but not when you open the departures view).
(And I don't see the update in Google Play yet but I suppose this will just take a couple of hours)

schildbach:
It's live on the test channel already.

I'll look into the departures issue soon, I expect at least one quick follow-up release anyway. One particular thing I need to monitor is which devices will maybe choke on the Java streaming API (which wasn't used in PTE up to now).

traines-source:
Ah I didn't realize streams are such an issue, even though being on Java 8. Haven't done any Android development in a long time... Then we should just have replaced them with plain old Java I guess.

schildbach:
Well let's see. It would be heartbreaking to see your streams code being converted to old loops.

schildbach:
I just verified Android 8.1 works fine with streams. So I assume 8.0 does, too, because it's just a minor version bump between the two. So only Android 7/7.1 remains unknown for now.

traines-source:
Well according to the docs it should work starting API level 24, so Android 7.0 should be fine, right? I was rather worried about other projects that are using PTE and that maybe haven't yet upgraded to minSdk 24. 
Another topic: In https://github.com/public-transport/db-vendo-client/issues/10 it was discovered that the some rate limit is at 60 requests/minute. If it's IP-based, it shouldn't be a problem for Oeffi, but we couldn't yet exclude that it is solely based on the `X-Correlation-ID` (or some other) header, in which case all Oeffi users together might reach it (not sure how active your users are :D). In this case and if you push another release anyways, one could change this from `"null"` to randomly generated UUIDs, to better mimic the behavior of DB Navigator. What do you think?

schildbach:
Does the API limit include station board requests? If yes, a single user could hit that limit if scrolling a lot.

Regarding `X-Correlation-ID`, sure, let's mimic DB Navigator. What format of UUIDs do they use?

schildbach:
Answering my own Q: `0564dcdf-edbf-4412-a147-7299f6481470_64466773-556f-4aa8-b128-0948c4d60887

schildbach:
@traines-source see 5522113051377aebe1160ff9c63a1f04bcc96e75

traines-source:
Jup thats exactly how marudor (and I in db-vendo-client now) does it, too :+1:

schildbach:
I found an incompatibility: `java.time` API isn't available until Android 8.

Would you like to have a look at this PR, which converts it to old-style DateFormat?

https://github.com/schildbach/public-transport-enabler/pull/621

schildbach:
@traines-source Another Java 8-ism: https://github.com/schildbach/public-transport-enabler/pull/623

Would you like to have a look at the proposed replacement?

PR #619, opened 2025-01-10
HttpClient: allow passing in custom Accept header

schildbach:


traines-source:
haven't tested it, but LGTM, should have the same effect :+1:

schildbach:
Merged.

PR #623, opened 2025-01-15
DB: replace usage of String.join() with java.util.StringJoiner

schildbach:
String.join() isn't available until Android 8.

### DIFF
--- a/src/de/schildbach/pte/DbProvider.java
+++ b/src/de/schildbach/pte/DbProvider.java
@@ -36,6 +36,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.StringJoiner;
 import java.util.TimeZone;
 import java.util.UUID;
 import java.util.regex.Matcher;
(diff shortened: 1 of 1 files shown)


traines-source:
If you're sure that the double colon operator (`::`), which AFAIK is also Java 8, is available in Android SDK 24... – at least `StringJoiner` seems to be, even though it's also Java 8.
Shouldn't Android Studio tell you when you're using stuff that is not supported on your minSdk level? (I've only built on the command line.) Or should we maybe just compile with Java 7 (and yes, remove the streams) instead of finding these things one by one?

schildbach:
According to the crash logs, this is the only remaining issue. And it's actually happening quite rarely – only about 1.2% of devices are still using Android 7.x, they need to be using DB provider, and they need to parse a connection with messages. I expect Öffi to require API 26 by end of this year or so.

As for method references (double colon op), according to https://developer.android.com/studio/write/java8-support they're supported on any API level.

traines-source:
in that case LGTM :+1: (and sorry for the oversight)

schildbach:
No worries, and thanks for your review! Merged.

Ahz9619 (changes requested):
****

PR #674, opened 2025-11-17
RMV: new provider for the Rhine-Main metropolitan region

schildbach:
This RMV provider used to exist but was merged with the NVV provider via 0b17a9eb4779353e9d458ac963a5a1f251e86d75, because both seemed to yield the same data. From now on, they are separate again.


schildbach:
Merged, but without the styles for now.

Issue #676, opened 2025-11-26
Jitpack build fails


woheller69:
https://jitpack.io/com/github/schildbach/public-transport-enabler/master-ded67c8e3c-1/build.log

> Task :compileJava FAILED

   > Could not resolve com.google.guava:guava:33.5.0-android.

woheller69:
It turns out that starting with Guava 33.0.0, Google switched to [Gradle variant-aware publishing ](https://github.com/google/guava/releases/tag/v33.0.0), which causes ambiguity errors in plain Java projects unless the JVM environment is explicitly declared. 

To fix this and keep compatibility with both standard Java and Android (API ≥ 26), the following changes are needed:

```implementation 'com.google.guava:guava:33.5.0-jre'``` instead of ```com.google.guava:guava:33.5.0-android```

Help Gradle disambiguate the variant by declaring the standard JVM environment:

```
java {
    toolchain {
        languageVersion = JavaLanguageVersion.of(8)
    }
}

configurations.configureEach {
    if (name.contains('Classpath')) {
        attributes {
            attribute(Attribute.of('org.gradle.jvm.environment', String), 'standard-jvm')
        }
    }
}
```

**Comment by schildbach** [comment]
It uses Debian Gradle 4.4.1. Can Jitpack be configured to use that?

**Comment by woheller69** [comment]
This is what Jitpack uses:
```grep: gradle/wrapper/gradle-wrapper.properties: No such file or directory
Generating Gradle wrapper v 6.7.1
Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF-8 -Dhttps.protocols=TLSv1.2

Welcome to Gradle 4.8.1!
```

In order to define which Gradle to use you should add a gradle/wrapper directory with a gradle-wrapper.properties file.
and add e.g. 
```distributionUrl=https://services.gradle.org/distributions/gradle-4.4.1-bin.zip```

It seems the core issue is that JitPack uses newer Gradle, which enforces variant selection for new Guava.
So you can either downgrade Guava or add the definition which Gradle to use.

PR #678, opened 2025-11-27
Add gradle-wrapper.properties for Gradle 4.4.1

woheller69:
#676
With that at least build on Jitpack will work and use the same Gradle version as CI

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -50,3 +50,15 @@ compileTestJava {
 test {
     exclude 'de/schildbach/pte/live/**'
 }
+
+publishing {
+    publications {
+        maven(MavenPublication) {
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -50,3 +50,15 @@ compileTestJava {
...
+    publications {
+        maven(MavenPublication) {
+            groupId = project.group
+            artifactId = archivesBaseName
+            version = project.version
+
+            from components.java
+        }
+    }
+}
```
schildbach: What is this good for? How is it related to the topic of this PR?

woheller69: This is needed for maven publish. Without this block Jitpack does not find the artifact.

PR #711, opened 2026-05-01
AVV_Aachen: Additional city abbreviations

thtomate:
As already hinted in #554, there is more mess with station names. Here is another fix: DN and JÜL are also used to abbreviate cities

(POI and Address tests are failing, but they were failing before, see #712)

### DIFF
--- a/src/de/schildbach/pte/AvvAachenProvider.java
+++ b/src/de/schildbach/pte/AvvAachenProvider.java
@@ -47,25 +47,39 @@ public class AvvAachenProvider extends AbstractHafasClientInterfaceProvider {
         setApiAuthorization(apiAuthorization);
     }
 
-    private static final String[] PLACES = { "AC", "Aachen" };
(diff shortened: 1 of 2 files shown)


schildbach (commented):
This `splitStationName()` implementation starts to feel a bit convoluted to me. Especially the now duplicated list of places I'd like to avoid.

I wonder if `PLACES` should perhaps be converted into a map, and then both the keys and values are matched in the loop below but the value always used. Wdyt?

And just for reference, both `RmvProvider` and `NvvProvider` solve a similar problem with abbreviated city names. I don't like their current solution either, but maybe whatever we improve here can then be carried over to there.

thtomate:
Yes, you are absolutely right— joking. Of course, a more generic solution is preferred if the need is also at other places.
However, I think about contacting them, as the situation with station names is getting worse everytime I look into it: Now, we have dots as separator for some stations outside AVV area…

PR #723, opened 2026-05-31
NetworkProvider: remove parameter `equivs` from `queryDepartures()`

schildbach:
The concept is badly defined between providers and a headache to maintain. On modern APIs with hierarchical locations (station, stop, position), you'd better query on a "higher level".

This also removes the `stationBoardCanDoEquivs` property from `AbstractHafasLegacyProvider`.

### DIFF
--- a/src/de/schildbach/pte/AbstractEfaProvider.java
+++ b/src/de/schildbach/pte/AbstractEfaProvider.java
@@ -1472,14 +1472,14 @@ public abstract class AbstractEfaProvider extends AbstractNetworkProvider {
 
     @Override
     public QueryDeparturesResult queryDepartures(final String stationId, final @Nullable Date time,
-            final int maxDepartures, final boolean equivs) throws IOException {
+            final int maxDepartures) throws IOException {
         requireNonNull(stationId);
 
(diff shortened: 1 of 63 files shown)


schildbach:
I checked Öffi and Transportr don't use `equivs` (it's always false).

Öffi NG has evolved the boolean into an enum `EquivalentStationsMode` with three values, but uses only two of them: `USE_META` and `META_IF_SAME_NAME` – not sure what they mean though.

Anyone aware of any other relevant consumers?

traines-source:
The concept being a fuzzy one IMO won't change with the removal of the flag, because still different providers will consider different things the same station. I do think that there might be more indirect users of this library, although I cannot name any concrete ones. Also I would have thought that ignoring the flag where it's not possible to implement or do a best effort implementation (like for Transitous with a simple radius=500 parameter) would have been simpler and less disruptive than removing it altogether.

1Maxnet1:
> Anyone aware of any other relevant consumers?

There is https://github.com/you-apps/TransportYou which recently switched to the Öffi NG fork, if that's relevant enough.

